PR #12, opened 2021-01-12
Create Architectury Item Properties

OroArmor:


OroArmor:
Any updates on this? I know you want a rewrite, but it would be nice to have no expect platforms in my code

shedaniel:
I am thinking if we can change all of these into an interface you can implement into `ItemExtension` that is in the API now. However with this way we will need to go over fabric api and implement the features with mixins ourselves.

PR #13, opened 2021-01-15
Add AbstractRecipeSerializer

shedaniel:
We ASM ourselves this round, our code can't get worse.


MaxNeedsSnacks:
Seems to work well in testing on Forge; testing on Fabric shouldn't be needed because we're not doing anything, so I'll be merging this into the main branch ^^

PR #22, opened 2021-01-25
Add GuiEvent.SET_SCREEN

MaxNeedsSnacks:
Could probably use an implementation in the test mod, but from testing pretty much this exact impl in https://github.com/KubeJS-Mods/KubeJS-UI/pull/1, we should be good to go.

### DIFF
--- a/common/src/main/java/me/shedaniel/architectury/event/events/GuiEvent.java
+++ b/common/src/main/java/me/shedaniel/architectury/event/events/GuiEvent.java
@@ -50,44 +50,44 @@ public interface GuiEvent {
     Event<ScreenInitPost> INIT_POST = EventFactory.createLoop(ScreenInitPost.class);
     Event<ScreenRenderPre> RENDER_PRE = EventFactory.createInteractionResult(ScreenRenderPre.class);
     Event<ScreenRenderPost> RENDER_POST = EventFactory.createInteractionResult(ScreenRenderPost.class);
-
+    
     /**
      * Invoked during Minecraft#setScreen, equivalent to forge's {@code GuiOpenEvent}.
      */
-    Event<SetScreenEvent> SET_SCREEN = EventFactory.createInteractionResultHolder();
-
-    @Environment(EnvType.CLIENT)
-    interface SetScreenEvent {
-        InteractionResultHolder<Screen> modifyScreen(Screen screen);
-    }
-
+    Event<SetScreen> SET_SCREEN = EventFactory.createInteractionResultHolder();
+    
     @Environment(EnvType.CLIENT)
     interface RenderHud {
         void renderHud(PoseStack matrices, float tickDelta);
     }
-
+    
     @Environment(EnvType.CLIENT)
     interface DebugText {
         void gatherText(List<String> strings);
     }
-
+    
     @Environment(EnvType.CLIENT)
     interface ScreenInitPre {
         InteractionResult init(Screen screen, List<AbstractWidget> widgets, List<GuiEventListener> children);
     }
-
+    
     @Environment(EnvType.CLIENT)
     interface ScreenInitPost {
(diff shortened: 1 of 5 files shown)


MaxNeedsSnacks:
I should probably mention that this PR also contains a patch to move initialisation of the initial screen (either a title screen or a multiplayer connect screen) to **after** the resource manager has been fully reloaded for the first time, a fix that [Forge implements by default](https://github.com/MinecraftForge/MinecraftForge/blob/1.16.x/patches/minecraft/net/minecraft/client/Minecraft.java.patch). As stated in said Forge patch this fixes [MC-145102](https://bugs.mojang.com/browse/MC-145102) and also adds capabilities for mods to wrap and override the main menu using custom resources (again, like KubeJS-UI, which is why I discovered that patch in the first place)

forge/src/main/java/me/shedaniel/architectury/event/forge/EventHandlerImplClient.java
```diff
@@ -115,59 +116,69 @@ public static void event(ClientChatReceivedEvent event) {
...
-    
+
+    @SubscribeEvent
+    public static void event(GuiOpenEvent event) {
+        InteractionResultHolder<Screen> result = GuiEvent.SET_SCREEN.invoker().modifyScreen(event.getGui());
+        if (result.getResult() == InteractionResult.FAIL) {
+            event.setCanceled(true);
+            return;
+        }
+        event.setGui(result.getObject());
```
shedaniel: Only apply if SUCCESS

common/src/main/java/me/shedaniel/architectury/event/events/GuiEvent.java
```diff
@@ -49,32 +50,42 @@
...
+    @Environment(EnvType.CLIENT)
+    interface SetScreenEvent {
+        InteractionResultHolder<Screen> modifyScreen(Screen screen);
+    }
+
     @Environment(EnvType.CLIENT)
     interface RenderHud {
         void renderHud(PoseStack matrices, float tickDelta);
     }
-
```
shedaniel: can you format this

forge/src/main/java/me/shedaniel/architectury/event/forge/EventHandlerImplClient.java
MaxNeedsSnacks: What should the analogue be on fabric? Should we also only change on SUCCESS?

shedaniel: Yep

fabric/src/main/java/me/shedaniel/architectury/mixin/fabric/client/MixinMinecraft.java
```diff
@@ -59,4 +77,78 @@ private void rightClickAir(CallbackInfo ci, InteractionHand var1[], int var2, in
...
+    
+    @Redirect(
+            method = "<init>",
+            at = @At(value = "INVOKE", target = "Lnet/minecraft/client/Minecraft;setScreen(Lnet/minecraft/client/gui/screens/Screen;)V"),
+            slice = @Slice(
+                    from = @At(value = "INVOKE", target = "Lnet/minecraft/client/Minecraft;resizeDisplay()V"),
+                    to = @At(value = "INVOKE", target = "Lnet/minecraft/client/gui/screens/LoadingOverlay;registerTextures(Lnet/minecraft/client/Minecraft;)V")
+            )
+    )
+    public void minecraftWhy(Minecraft mc, Screen screen) {
```
MaxNeedsSnacks, line 128: This is the patch I mentioned in my comment that moves setting the initial screen to after resource reloading has occurred.

MaxNeedsSnacks:
Couldn't find any conflicting mixins on GitHub so I'm pretty sure this is good to go.
Let me test it with KubeJS-UI real quick...

PR #23, opened 2021-01-25
Fix mojank in EffectInstance

MaxNeedsSnacks:
This is another one of those mojank patches by forge I discovered while porting KubeJS-UI

Might be useful to have in architectury to get rid of some vanilla bugs devs moving from exclusively Forge-based setups may encounter

### DIFF
--- a/fabric/src/main/java/me/shedaniel/architectury/mixin/fabric/client/MixinEffectInstance.java
+++ b/fabric/src/main/java/me/shedaniel/architectury/mixin/fabric/client/MixinEffectInstance.java
@@ -1,0 +1,38 @@
+package me.shedaniel.architectury.mixin.fabric.client;
+
+import com.mojang.blaze3d.shaders.Program;
+import net.minecraft.client.renderer.EffectInstance;
+import net.minecraft.resources.ResourceLocation;
+import net.minecraft.server.packs.resources.ResourceManager;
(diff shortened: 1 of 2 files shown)


shedaniel:
This is going to break some fabric mods depending on this broken behaviour, i.e. the armour texture is limited to the minecraft namespace, and you know what people did? They just add their texture to the minecraft namespace to circumvent the issue.

MaxNeedsSnacks:
How would you say we should circumvent that? Obviously this isn't intentional on Mojang's end and the original format is still supported, i.e. mods adding things to the mc namespace SPECIFICALLY won't break with this, but this (and similar issues like it) will probably be a problem a fair few modders migrating from Forge will encounter on their way.

shedaniel:
It is _possible_ to do both formats (by detecting via the ResourceManager), but it is annoying.

MaxNeedsSnacks:
The thing I'm wondering is: Would this actually break with anything?
The original code throws a ResourceLocationException when given something like `fabricmod:mymodel`, so no mod should be using that in production either (and as you said, they don't and instead would instead just use a string id like `mymodel`, which under this new impl would still be put under the `minecraft` namespace).

shedaniel:
Right indeed, this should be fine then, we can probably merge this now.

PR #62, opened 2021-03-11
Rewrite the GuiEvent.SET_SCREEN mixin to fix #35

### LINKED ISSUE #35: Conflict with Voxelmap
[crash-2021-02-01_11.54.47-client.txt](https://github.com/architectury/architectury-api/files/5901771/crash-2021-02-01_11.54.47-client.txt)

shedaniel:


MaxNeedsSnacks (approved):
Continuing our discussion from Discord, I would argue that mod intercompatibility is more important here than moving main menu init to post-resource reload, at least by default. The initial reason for this fix was that KubeJS UI loads UI definitions from client scripts and then replaces Screens on init if they have a matching KJS UI Wrapper. This wasn’t possible in fabric / vanilla since by default, the main menu already gets initialised before KJS UI has a chance to replace it, but I guess we’ll just have to find a different fix for that at some other point down the line

PR #64, opened 2021-03-21
Add EntityHooks.fromCollision to retrieve an entity from block collision

MaxNeedsSnacks:
Tiny thing I encountered while working on another mod, *should* work without issue, still making this a PR so others can test before merging


fabric/src/main/java/me/shedaniel/architectury/mixin/fabric/MixinEntityCollisionContext.java
```diff
@@ -0,0 +1,32 @@
...
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.injection.At;
+import org.spongepowered.asm.mixin.injection.Inject;
+import org.spongepowered.asm.mixin.injection.callback.CallbackInfo;
+
+@Mixin(EntityCollisionContext.class)
+@Implements(@Interface(iface = EntityHooksImpl.CollisionContextExtension.class, prefix = "arch$"))
+public abstract class MixinEntityCollisionContext implements CollisionContext {
+    
+    private Entity arch$entity = null;
```
shedaniel: use weak reference and unique

PR #69, opened 2021-04-01
Possible fixes for BE syncing on Forge

MaxNeedsSnacks:
Experimenting around with different ways to fix BE syncing issues on Forge
(originally discovered in FTB Quests that BEs don't seem to properly sync their data by default)

### DIFF
--- a/common/src/main/java/me/shedaniel/architectury/extensions/BlockEntityExtension.java
+++ b/common/src/main/java/me/shedaniel/architectury/extensions/BlockEntityExtension.java
@@ -20,6 +20,7 @@
 package me.shedaniel.architectury.extensions;
 
 import me.shedaniel.architectury.hooks.BlockEntityHooks;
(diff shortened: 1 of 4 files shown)


LatvianModder:
nice

MaxNeedsSnacks:
also... nice

MaxNeedsSnacks:
nice

shedaniel:
is this intended? you are supposed to call that super saveMetadata method in your own saveClientData method

MaxNeedsSnacks:
Since saveMetadata is a private method that's only used in base BlockEntity, I'm guessing that would be getUpdateTag, however if that's the case I don't 100% get what to do with the input tag parameter in saveClientData on Forge as modders would obviously want to use the result of `super.getUpdateTag()` instead to preserve their superclass's data as well.

Also is there any case where a BE would **not** want its super's metdata added on top, out of curiosity?

shedaniel:
Actually, not `saveMetadata`, I meant the normal `save`.

PR #72, opened 2021-04-10
Adds hook to tick armor and return custom equipment slots

### LINKED ISSUE #39: Item Armor Tick
Tick the item when worn as armor.

shedaniel:
Adds `ItemExtension` as the same style as `BlockEntityExtension`, with a method called `tickArmor` and one called `getCustomEquipmentSlot`.

Close #39

### DIFF
--- a/common/src/main/java/me/shedaniel/architectury/extensions/ItemExtension.java
+++ b/common/src/main/java/me/shedaniel/architectury/extensions/ItemExtension.java
@@ -24,6 +24,9 @@ import net.minecraft.world.entity.player.Player;
 import net.minecraft.world.item.ItemStack;
 import org.jetbrains.annotations.Nullable;
 
+/**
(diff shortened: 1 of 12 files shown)


common/src/main/java/me/shedaniel/architectury/extensions/ItemExtension.java
```diff
@@ -0,0 +1,47 @@
...
+ */
+
+package me.shedaniel.architectury.extensions;
+
+import net.minecraft.world.entity.EquipmentSlot;
+import net.minecraft.world.entity.player.Player;
+import net.minecraft.world.item.ItemStack;
+import org.jetbrains.annotations.Nullable;
+
+public interface ItemExtension {
```
Juuxel, line 30: IMO this should document how to use this (= "implement this on your item thanks")

PR #75, opened 2021-04-14
Some more (mostly Entity-based) Events

MaxNeedsSnacks:
This adds cross-platform events based on the following Forge callbacks:
- FillBucketEvent
- BlockEvent.FarmlandTrampleEvent
- EntityEvent.EnteringChunk
- LivingSpawnEvent.CheckSpawn

as well as test mods for every one of them except CHECK_SPAWN


common/src/main/java/me/shedaniel/architectury/event/events/EntityEvent.java
```diff
@@ -67,4 +79,12 @@
     interface PlaceBlock {
         InteractionResult placeBlock(Level world, BlockPos pos, BlockState state, @Nullable Entity placer);
     }
+    
+    interface EnterChunk {
+        void enterChunk(Entity entity, int chunkX, int chunkZ, int prevX, int prevZ);
+    }
+    
+    interface CheckSpawn {
+        InteractionResult canSpawn(Entity entity, LevelAccessor world, double x, double y, double z, MobSpawnType type, @Nullable BaseSpawner spawner);
```
LatvianModder: Shouldn't this be LivingEntity?

MaxNeedsSnacks: Forge's event only gives us an Entity (despite explicitly only being designed for Mobs), but if people want to, they can cast it themselves

PR #78, opened 2021-04-20
Quilt Support

shedaniel:
Update Architectury Plugin to 3.1, added support for quilt detection, see architectury/architectury-transformer#3 for more information.

Architectury#getModLoader and Platform#getModLoader have been deprecated and will still return "fabric" on a quilt environment.

It is currently unknown if QuiltMC/quilt-loader would relocate its packages in the near future, the Quilt project has not been released yet, the current code is subject to changes, yet, there is no good way to detect quilt other than checking the presence of its implementation of the Fabric Loader.

This PR will hold off until everything is finalised.


shedaniel:
Outdated, we will think about this again in the future

PR #81, opened 2021-05-10
Add FuelRegistry

### LINKED ISSUE #37: Fuel API
Remember, forge uses an ItemStack-aware method in IForgeItem, but in fabric you just add to the map using the api.

shedaniel:
Close #37 

### DIFF
--- a/forge/src/main/java/me/shedaniel/architectury/registry/fuel/forge/FuelRegistryImpl.java
+++ b/forge/src/main/java/me/shedaniel/architectury/registry/fuel/forge/FuelRegistryImpl.java
@@ -1,0 +1,61 @@
+/*
+ * This file is part of architectury.
+ * Copyright (C) 2020, 2021 architectury
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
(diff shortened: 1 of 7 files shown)


forge/src/main/java/me/shedaniel/architectury/registry/fuel/forge/FuelRegistryImpl.java
```diff
@@ -0,0 +1,61 @@
...
+import net.minecraft.world.level.ItemLike;
+import net.minecraftforge.event.furnace.FurnaceFuelBurnTimeEvent;
+import net.minecraftforge.eventbus.api.SubscribeEvent;
+
+public class FuelRegistryImpl {
+    private static final Object2IntMap<ItemLike> ITEMS = new Object2IntLinkedOpenHashMap<>();
+    
+    static {
+        EventBuses.onRegistered(ArchitecturyForge.MOD_ID, bus -> {
+            bus.register(FuelRegistryImpl.class);
```
MaxNeedsSnacks: `FurnaceFuelBurnTimeEvent` is fired on the `MinecraftForge.EVENT_BUS`, not the mod bus, see https://github.com/MinecraftForge/MinecraftForge/blob/b6ce4fb066eb886a305f0678649c33d6f8ba4f76/src/main/java/net/minecraftforge/event/furnace/FurnaceFuelBurnTimeEvent.java#L42

MaxNeedsSnacks (changes requested):

MaxNeedsSnacks (approved):

Trikzon (approved):
Looks good to me and is pretty close to what I have done myself. Only real difference is passing an `ItemStack` instead of an `Item` to `get()`. I'm not sure why a stack would hold a different burn time than its item, but it's nice to see it supported.

MaxNeedsSnacks:
That is to futureproof our api for stack-aware approaches down the line, should FAPI add support for them eventually

A situation where a stack may have a different burn time than the base item would be something like Extra Utilities‘ Redstone Coil, which would extract from the player‘s power network to heat a furnace

shedaniel:
Plus forge needs it anyways.

PR #84, opened 2021-05-14
Add an easy way to register Villager and Wandering Trader trades

### LINKED ISSUE #70: Villager Trade Event
Should be relatively trivial to implement, even if a corresponding event doesn't seem(?) to exist on the fabric side.

That said, we should probably also look into utility functions to make adding Villager trades easier than the vanilla way, which is of course, janky as all hell once again.

canitzp:
As said in #70 this is a very trivial implimentation for Trade registering, but it makes the whole thing a lot easier.
Also when the forge modloader is used, the corresponding events (VillagerTradesEvent & WandererTradesEvent) are used.


fabric/src/main/java/me/shedaniel/architectury/registry/fabric/TradeRegistryImpl.java
```diff
@@ -0,0 +1,31 @@
+package me.shedaniel.architectury.registry.fabric;
+
+import net.minecraft.world.entity.npc.VillagerProfession;
+import net.minecraft.world.entity.npc.VillagerTrades;
+import org.apache.commons.lang3.ArrayUtils;
+import org.apache.commons.lang3.math.NumberUtils;
+
+public class TradeRegistryImpl {
```
shedaniel: This needs to use the fabric trade registry, or else it will unregister on their own (Not our fault technically, their api is really really awful)

shedaniel (changes requested):

fabric/src/main/java/me/shedaniel/architectury/registry/fabric/TradeRegistryImpl.java
canitzp: I didn't know that it existed, but I changed it

common/src/main/java/me/shedaniel/architectury/registry/TradeRegistry.java
```diff
@@ -0,0 +1,38 @@
...
+import me.shedaniel.architectury.annotations.ExpectPlatform;
+import net.minecraft.world.entity.npc.VillagerProfession;
+import net.minecraft.world.entity.npc.VillagerTrades;
+
+public class TradeRegistry {
+    
+    private TradeRegistry(){}
+    
+    /**
+     * Register a Trade (ItemListing) for a villager by it's profession and level.
```
shedaniel: This could use a {@link}

forge/src/main/java/me/shedaniel/architectury/registry/forge/TradeRegistryImpl.java
```diff
@@ -0,0 +1,65 @@
...
+import net.minecraftforge.fml.common.Mod;
+import org.apache.commons.lang3.ArrayUtils;
+
+import java.util.*;
+
+@Mod.EventBusSubscriber(modid = ArchitecturyForge.MOD_ID)
+public class TradeRegistryImpl {
+    
+    private static final Map<VillagerProfession, Int2ObjectMap<List<VillagerTrades.ItemListing>>> TRADES_TO_ADD = new HashMap<>();
+    private static final Int2ObjectMap<List<VillagerTrades.ItemListing>> WANDERER_TRADES_TO_ADD = new Int2ObjectOpenHashMap<>();
```
shedaniel: This could use a Boolean2ObjectMap

common/src/main/java/me/shedaniel/architectury/registry/TradeRegistry.java
Juuxel:
```suggestion
     * Register a trade ({@link VillagerTrades.ItemListing}) for a villager by its profession and level.
```

common/src/main/java/me/shedaniel/architectury/registry/TradeRegistry.java
```diff
@@ -0,0 +1,38 @@
...
+     *
+     * @param profession The Profession the villager needs to have this trade.
+     * @param level The level the villager needs. Range is 1 to 5 with 1 being the lowest level and 5 the highest.
+     * @param trades The trades to add to this profession at the specified level.
+     *               
+     * When the Modloader is forge the {@code VillagerTradesEvent} event is used.
```
Juuxel:
```suggestion
     * When the mod loader is Forge, the {@code VillagerTradesEvent} event is used.
```

common/src/main/java/me/shedaniel/architectury/registry/TradeRegistry.java
```diff
@@ -0,0 +1,38 @@
...
+     */
+    @ExpectPlatform
+    public static void registerVillagerTrade(VillagerProfession profession, int level, VillagerTrades.ItemListing... trades){
+        throw new AssertionError();
+    }
+    
+    /**
+     * Register a Trade (ItemListing) to a wandering trader by its rarity.
```
Juuxel:
```suggestion
     * Register a trade ({@link VillagerTrades.ItemListing}) to a wandering trader by its rarity.
```

common/src/main/java/me/shedaniel/architectury/registry/TradeRegistry.java
```diff
@@ -0,0 +1,38 @@
...
+     *
+     * @param rare Is this a rare trade. Rare trades only have a five times lower chance of being used.
+     * @param trades The trades to add to the wandering trader.
+     *
+     * When the Modloader is forge the {@code WandererTradesEvent} event is used.
```
Juuxel:
```suggestion
     * When the mod loader is Forge, the {@code WandererTradesEvent} event is used.
```

common/src/main/java/me/shedaniel/architectury/trade/SimpleTrade.java
```diff
@@ -0,0 +1,34 @@
...
+
+import net.minecraft.world.entity.Entity;
+import net.minecraft.world.entity.npc.VillagerTrades;
+import net.minecraft.world.item.ItemStack;
+import net.minecraft.world.item.trading.MerchantOffer;
+import org.jetbrains.annotations.Nullable;
+
+import java.util.Random;
+
+public class SimpleTrade implements VillagerTrades.ItemListing {
```
Juuxel: IMO this class should be documented (including why it's needed)

forge/src/main/java/me/shedaniel/architectury/registry/forge/TradeRegistryImpl.java
Juuxel: `Boolean2ObjectMap` would make more sense here (as the key is pretty much the value of `rare`)

Juuxel (changes requested):

forge/src/main/java/me/shedaniel/architectury/registry/forge/TradeRegistryImpl.java
canitzp: A Boolean2ObjectMap does not exist within fastutil

Juuxel: Ah, I'd just use two list fields then. The outer map isn't really needed.

common/src/main/java/me/shedaniel/architectury/trade/SimpleTrade.java
canitzp: Done

forge/src/main/java/me/shedaniel/architectury/registry/forge/TradeRegistryImpl.java
canitzp: done

common/src/main/java/me/shedaniel/architectury/registry/TradeRegistry.java
```diff
@@ -0,0 +1,38 @@
...
+    /**
+     * Register a trade ({@link VillagerTrades.ItemListing}) to a wandering trader by its rarity.
+     *
+     * @param rare Is this a rare trade. Rare trades only have a five times lower chance of being used.
+     * @param trades The trades to add to the wandering trader.
+     *
+     * When the mod loader is Forge, the {@code WandererTradesEvent} event is used.
+     */
+    @ExpectPlatform
+    public static void registerTradeForWanderer(boolean rare, VillagerTrades.ItemListing... trades){
```
Juuxel: I guess this should use `WanderingTrader` instead of `Wanderer` (I think that's a bad Forge name, actually?)

Juuxel (approved):
Looks good apart from a minor naming thing.

common/src/main/java/me/shedaniel/architectury/registry/TradeRegistry.java
shedaniel: Or try to be consistent with the method above?
registerWanderingTraderTrade, it may be a mouthful

forge/src/main/java/me/shedaniel/architectury/registry/forge/TradeRegistryImpl.java
```diff
@@ -0,0 +1,65 @@
...
+@Mod.EventBusSubscriber(modid = ArchitecturyForge.MOD_ID)
+public class TradeRegistryImpl {
+    
+    private static final Map<VillagerProfession, Int2ObjectMap<List<VillagerTrades.ItemListing>>> TRADES_TO_ADD = new HashMap<>();
+    private static final List<VillagerTrades.ItemListing> WANDERER_TRADER_TRADES_GENERIC = new ArrayList<>();
+    private static final List<VillagerTrades.ItemListing> WANDERER_TRADER_TRADES_RARE = new ArrayList<>();
+    
+    public static void registerVillagerTrade(VillagerProfession profession, int level, VillagerTrades.ItemListing... trades) {
+        if(level < 1 || level > 5){
+            throw new RuntimeException("Villager Trade level has to be between 1 and 5!");
```
shedaniel: IllegalArgumentException, and print the current invalid value (Change this in the fabric one as well)

shedaniel (approved):
Other than some small nitpacks, this is alright, I'll fix the code style and licenses on merge

common/src/main/java/me/shedaniel/architectury/registry/TradeRegistry.java
shedaniel: Somehow my previous draft reviews slipped through 😅

common/src/main/java/me/shedaniel/architectury/registry/TradeRegistry.java
canitzp: done

common/src/main/java/me/shedaniel/architectury/registry/TradeRegistry.java
canitzp: already done

forge/src/main/java/me/shedaniel/architectury/registry/forge/TradeRegistryImpl.java
canitzp: already done

forge/src/main/java/me/shedaniel/architectury/registry/forge/TradeRegistryImpl.java
canitzp: done

canitzp:
Maybe it is a idea to add the level check back, but only check if the passed level is greater or equal 1? I added this bounds check, because I added a level 0 in the test mod first and after half an hour wasted debugging, I realised that the lowest level is 1 and not 0 🤦 .
I understand that adding level is possible, so the upper bound is removed

shedaniel:
That sounds plausible, I will do that.

MaxNeedsSnacks (approved):

PR #87, opened 2021-05-17
Fix TradeRegistryImpl classes. With the private access modifier the a…

canitzp:
…rchitectury transformer throws an AssertError while loading the game. After changing them from private to public, the game loads like normal.

I noticed this error, while testing my new pr.

### DIFF
--- a/fabric/src/main/java/me/shedaniel/architectury/registry/trade/fabric/TradeRegistryImpl.java
+++ b/fabric/src/main/java/me/shedaniel/architectury/registry/trade/fabric/TradeRegistryImpl.java
@@ -26,7 +26,7 @@ import net.minecraft.world.entity.npc.VillagerTrades;
 import java.util.Collections;
 
 public class TradeRegistryImpl {
-    private static void registerVillagerTrade0(VillagerProfession profession, int level, VillagerTrades.ItemListing... trades) {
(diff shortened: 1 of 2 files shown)


fabric/src/main/java/me/shedaniel/architectury/registry/trade/fabric/TradeRegistryImpl.java
```diff
@@ -26,7 +26,7 @@
 import java.util.Collections;
 
 public class TradeRegistryImpl {
-    private static void registerVillagerTrade0(VillagerProfession profession, int level, VillagerTrades.ItemListing... trades) {
+    public static void registerVillagerTrade0(VillagerProfession profession, int level, VillagerTrades.ItemListing... trades) {
```
MaxNeedsSnacks, line 29: Could you add @ApiStatus.Internal to both of these and the base method to let people know they *should not use this* (as it doesn't perform input validation)

Juuxel: The common method is private and platform impls can be assumed to be internal (IMO), so I don't think it's needed necessarily.

canitzp: I also don't think it is needed. When I look at other platform impl, none of them is marked as internal

PR #88, opened 2021-05-17
New utility hook for creating a entity spawn packet. before every mod…

canitzp:
… had to implement this in itself.

I implemented this in my mod Trainz and now decided to request the merge into architectury itself. The test mod is to check if everything works, which you can check by summoning the test_entity and looking if a minecart is rendered.

The test mod only works when #87 is merged.


testmod-common/src/main/java/me/shedaniel/architectury/test/TestMod.java
```diff
@@ -42,7 +45,9 @@ public static void initialize() {
         TestGameRules.init();
         TestTags.initialize();
         TestTrades.init();
-        if (Platform.getEnvironment() == Env.CLIENT)
+        if (Platform.getEnvironment() == Env.CLIENT) {
             TestKeybinds.initialize();
+            EntityRenderers.register(TestEntity.TYPE, MinecartRenderer<TestEntity>::new);
```
Juuxel, line 50: This will crash if the test mod is run on a dedicated server.

canitzp: I just fired one up and it doesn't, probably because of the side check two lines above

Juuxel: Huh, I would expect the class loader to crash because the method references `MinecartRenderer` (which is client-only), but the method reference might prevent that.

canitzp: Yeah the method is only references when it ever reaches that Class call, but it never does on a dedicated server, because of the client check.
I use this principle a lot, like I never use forge proxies, but just checking if the mod is loaded client or serverside and then go into a different method for handling the side. I needed to understand the bytecode first, but now it is very obvious for me.

common/src/main/java/me/shedaniel/architectury/networking/NetworkManager.java
```diff
@@ -74,6 +75,19 @@ public static boolean canPlayerReceive(ServerPlayer player, ResourceLocation id)
         throw new AssertionError();
     }
     
+    /**
+     * Easy to use utility method to create a entity spawn packet.
```
shedaniel:
```suggestion
     * Easy to use utility method to create an entity spawn packet.
```

common/src/main/java/me/shedaniel/architectury/networking/NetworkManager.java
```diff
@@ -74,6 +75,19 @@ public static boolean canPlayerReceive(ServerPlayer player, ResourceLocation id)
...
+     * This packet is needed everytime any mod adds a entity.
+     * The entity should override {@link Entity#getAddEntityPacket()} to point to this method!
+     *
+     * @param entity The entity which should be spawned. The 'this' keyword should do the trick all the time.
```
shedaniel:
```suggestion
     * @param entity The entity which should be spawned.
```
Don't think we need to cater to the absolute beginners.

fabric/src/main/java/me/shedaniel/architectury/networking/fabric/SpawnEntityPacket.java
```diff
@@ -0,0 +1,67 @@
...
+import net.minecraft.core.Registry;
+import net.minecraft.network.FriendlyByteBuf;
+import net.minecraft.network.protocol.Packet;
+import net.minecraft.resources.ResourceLocation;
+import net.minecraft.world.entity.Entity;
+import net.minecraft.world.entity.EntityType;
+
+import java.util.UUID;
+
+public class SpawnEntityPacket {
```
shedaniel, line 40: This class could use the architectury NetworkManager instead for easier maintainability.

canitzp: I'm not sure how to do this and I can't find any exaple

fabric/src/main/java/me/shedaniel/architectury/networking/fabric/SpawnEntityPacket.java
```diff
@@ -0,0 +1,67 @@
+package me.shedaniel.architectury.networking.fabric;
```
shedaniel, line 20: Need license

common/src/main/java/me/shedaniel/architectury/networking/NetworkManager.java
```diff
@@ -74,6 +75,19 @@ public static boolean canPlayerReceive(ServerPlayer player, ResourceLocation id)
         throw new AssertionError();
     }
     
+    /**
+     * Easy to use utility method to create an entity spawn packet.
+     * This packet is needed everytime any mod adds a entity.
+     * The entity should override {@link Entity#getAddEntityPacket()} to point to this method!
+     *
+     * @param entity The entity which should be spawned.
+     * @return It returns the ready to use packet, which can be returned by the {@link Entity#getAddEntityPacket()} method.
```
shedaniel:
```suggestion
     * @return The ready to use packet to spawn the entity on the client.
     * @see Entity#getAddEntityPacket()
```

common/src/main/java/me/shedaniel/architectury/networking/NetworkManager.java
```diff
@@ -74,6 +75,19 @@ public static boolean canPlayerReceive(ServerPlayer player, ResourceLocation id)
         throw new AssertionError();
     }
     
+    /**
+     * Easy to use utility method to create an entity spawn packet.
+     * This packet is needed everytime any mod adds a entity.
```
shedaniel:
```suggestion
     * This packet is needed everytime any mod adds a non-living entity.
```

shedaniel (changes requested):

shedaniel (approved):

MaxNeedsSnacks (approved):
Good to go in my eyes, moving this to last call!

PR #104, opened 2021-06-09
Implement phases for BiomeModifications on Forge

MaxNeedsSnacks:
I also added some basic Javadoc to explain the purpose of the four "phases".

Currently, the following priority holds:
1. Additions (HIGH)
2. Removals (NORMAL)
3. Replacements (LOW)
4. Post-Processing (LOWEST)

Feel free to give additional opinions on the matter before merging, as this is "only" medium-priority.

This feature, if applicable, will also be ported to 1.17 (we can just cherry pick the commits or merge later), however I think Forge *may* be getting an overhaul to their Biome system anyways so we may not need it like this in future versions

### DIFF
--- a/common/src/main/java/me/shedaniel/architectury/registry/BiomeModifications.java
+++ b/common/src/main/java/me/shedaniel/architectury/registry/BiomeModifications.java
@@ -37,16 +37,17 @@ import java.util.function.Predicate;
  * with the corresponding Forge EventPriority shown in brackets:
  *
  * <ol>
- *     <li>{@link #addProperties(Predicate, BiConsumer) Adding} new features to biomes. [HIGH]</li>
- *     <li>{@link #removeProperties(Predicate, BiConsumer)  Removing} existing features from biomes. [NORMAL]</li>
- *     <li>{@link #replaceProperties(Predicate, BiConsumer) Replacing} existing features with new ones. [LOW]</li>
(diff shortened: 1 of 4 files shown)


shedaniel (approved):

common/src/main/java/me/shedaniel/architectury/registry/BiomeModifications.java
```diff
@@ -27,6 +27,26 @@
...
+ * {@link net.fabricmc.fabric.api.biome.v1.ModificationPhase} enum.
+ *
+ * <p> The order in which these phases get processed is as follows,
+ * with the corresponding Forge EventPriority shown in brackets:
+ *
+ * <ol>
+ *     <li>{@link #addProperties(Predicate, BiConsumer) Adding} new features to biomes. [HIGH]</li>
+ *     <li>{@link #removeProperties(Predicate, BiConsumer)  Removing} existing features from biomes. [NORMAL]</li>
+ *     <li>{@link #replaceProperties(Predicate, BiConsumer) Replacing} existing features with new ones. [LOW]</li>
+ *     <li>Generic {@link #postProcessProperties(Predicate, BiConsumer) Post-Processing} of already modified biome features. [LOWEST]</li>
```
Juuxel: All these links should be `@linkplain` (plaintext) since the message isn't code.

Juuxel (approved):

PR #110, opened 2021-06-20
Particle Provider API

### LINKED ISSUE #93: [Request] Particle Factory Registration API
An API to register particle factories on the client. Necessary for any mod that wants to even approach custom particles.

**Forge:** [`ParticleFactoryRegisterEvent`](https://github.com/MinecraftForge/MinecraftForge/blob/1.16.x/src/main/java/net/minecraftforge/client/event/ParticleFactoryRegisterEvent.java) (Forge makes `ParticleManager.registerFactory` accessible)
**Fabric:** [`ParticleFactoryRegistry`](https://github.com/FabricMC/fabric/blob/1.16/fabric-particles-v1/src/main/java/net/fabricmc/fabric/api/client/particle/v1/ParticleFactoryRegistry.java)

~~Additionally, a utility class similar to [`FabricParticleTypes`](https://github.com/FabricMC/fabric/blob/1.16/fabric-particles-v1/src/main/java/net/fabricmc/fabric/api/particle/v1/FabricParticleTypes.java) would be appreciated, just so we don't need to extend `DefaultParticleType`.~~ nvm, seems like it's pretty much useless

Leo40Git:
Closes #93.

This API mirrors Fabric's equivalent API, meaning you can easily create animated particles that work on both sides!


Leo40Git:
I've added a test to the testmod (summon the test particle via `/particle` - I tried adding a block that randomly creates that particle but that failed).

Also, tested on Fabric 1.16.5.

common/src/main/java/me/shedaniel/architectury/registry/ParticleFactories.java
```diff
@@ -0,0 +1,55 @@
...
+        throw new AssertionError();
+    }
+
+    @ExpectPlatform
+    public static <T extends ParticleOptions> void register(ParticleType<T> type, PendingParticleProvider<T> constructor) {
+        throw new AssertionError();
+    }
+
+    @FunctionalInterface
+    public interface PendingParticleProvider<T extends ParticleOptions> {
```
shedaniel: DeferredParticleProvider

common/src/main/java/me/shedaniel/architectury/registry/ParticleFactories.java
```diff
@@ -0,0 +1,55 @@
...
+        List<TextureAtlasSprite> getSprites();
+    }
+
+    @ExpectPlatform
+    public static <T extends ParticleOptions> void register(ParticleType<T> type, ParticleProvider<T> provider) {
+        throw new AssertionError();
+    }
+
+    @ExpectPlatform
+    public static <T extends ParticleOptions> void register(ParticleType<T> type, PendingParticleProvider<T> constructor) {
```
shedaniel: Param: constructor -> provider
(fix this for the subprojects as well)

shedaniel (changes requested):

MaxNeedsSnacks (approved):

MaxNeedsSnacks:
Adding last call since Dan‘s review has been resolved

shedaniel (approved):

PR #115, opened 2021-06-30
Wrapper for BlockEntityType.Builder

shedaniel:
Since 1.17, `BlockEntityType$BlockEntitySupplier ` is now package-private, this adds a hook to allow constructing the block entity type without AW / Mixin.

### DIFF
--- a/common/src/main/java/dev/architectury/hooks/block/BlockEntityHooks.java
+++ b/common/src/main/java/dev/architectury/hooks/block/BlockEntityHooks.java
@@ -27,6 +27,7 @@ import net.minecraft.world.level.block.Block;
 import net.minecraft.world.level.block.entity.BlockEntity;
 import net.minecraft.world.level.block.entity.BlockEntityType;
 import net.minecraft.world.level.block.state.BlockState;
+import org.jetbrains.annotations.Nullable;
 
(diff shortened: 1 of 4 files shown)


canitzp (approved):

common/src/main/java/dev/architectury/hooks/block/BlockEntityHooks.java
```diff
@@ -19,18 +19,49 @@
...
+    public static class TypeBuilder<T extends BlockEntity> {
+        private final Constructor<? extends T> constructor;
+        private final Set<Block> validBlocks;
+        
+        private TypeBuilder(Constructor<? extends T> constructor, Set<Block> validBlocks) {
+            this.constructor = constructor;
+            this.validBlocks = validBlocks;
+        }
+        
+        public BlockEntityType<T> build(Type<?> type) {
```
Juuxel: Should have `@Nullable` (modded BEs don't generally have a DFU type and pass null, at least on fabric)

Juuxel (changes requested):

Juuxel (approved):

MaxNeedsSnacks (approved):

PR #116, opened 2021-06-30
Village trading

LLytho:
Hey,

as I wanted to add village trading to kubejs, max had the idea to implement it in architectury first. And because this is not the easiest decision how the interface should be look at the end a draft PR to discuss it. 

### Current state:
Currently there is only a way to add new trades through `SimpleTrade` or a custom implementation of `ItemListing`. While in the forge implementation you register trades with `registerVillagerTrade` which will be later on added through `VillagerTradesEvent` in fabric `registerVillagerTrade` directly modifies the `VillagerTrades.TRADES` map. 

### Possible problem:
With adding the ability to modify or remove trades you run into different execution orders for fabric and forge. So one of them should transformed into the same process as the other one. 

### Add/Register, Remove, Modify event (Idea credits to @MaxNeedsSnacks ): 
Split trading stuff into three events. 

#### Add
```java
public void onAddVillageTrades(AddFooEvent event) {// or direct arguments instead of a context
	event.getTrades(VillagerProfessions.FARMER, 3); // immutable list to get some trade information if needed?
	event.add(VillagerProfessions.FARMER, 1, (list) -> {
		list.Add(...);
		list.Add(...);
	});
}
```

#### Remove
```java
public void onRemoveVillageTrades(RemoveFooEvent event) {// or direct arguments instead of a context
	event.getTrades(VillagerProfessions.FARMER, 3); 
	
	// all three ways
	event.removeIf((list) -> {});
	event.removeIf(VillagerProfessions.FARMER, (list) -> {});
	event.removeIf(VillagerProfessions.FARMER, 3, (itemListing) -> { //  IItemListing 
		// `fromItem` is often called like this from mojang ...
		// On trades where no fromItem exists like `ItemForEmerald` getFromItem() should return `EMPTY`
		// IItemListing could implement `isXTrade()` to detect a specific ItemListing like `TreasureMapForEmeralds`
		return itemListing.getFromItem().getItem() == Item.DIAMOND; 
	});
	
	// other method ideas? 
}
```

#### Modify
```java
public void onModifyVillageTrades(ModifyFooEvent event) {// or direct arguments instead of a context
	const trades = event.getTrades(VillagerProfessions.FARMER, 3); 
	
	// not really sure what I like here more. Using just `trades` and filter the itemlisting by urself and call the setters or do it similiar to `removeIf`
}
```

### Maybe a problem
Because of `IItemListing` people could use `setFoo` in remove. 

## Some basic interface

```java
public interface IItemListing {  
    ItemStack getInput(); // decide if this is always strict to the input ITEM. If its `ItemsForEmerald` should it return EMPTY or Emerald? I like EMPTY more  
	
    void setInput(ItemStack itemStack);  

    // getInput too: Always returns a new ItemStack because of how 
    // mojang splits count and item into two different variables ... 🙃
    ItemStack getOutput();  
      
    void setOutput(ItemStack itemStack);  
      
    int getEmeraldCost();  
      
    void setEmeraldCost(int value);  
      
    int getMaxUses();  
      
    void setMaxUses(int value);  
      
    float getPriceMultiplier();  
      
    void setPriceMultiplier(float value);  
      
    IItemsAndEmeraldsToItems asItemsAndEmeraldsToItems();  
      
    IDyedArmorForEmeralds asDyedArmorForEmeralds();  
      
    IEffectItemForItemsAndEmeralds asEffectItemForItemsAndEmeralds();  
      
    IEmeraldForItems asEmeraldForItems();  
      
    IEmeraldsForVillagerTypeItem asEmeraldsForVillagerTypeItem();  
      
    IEnchantBookForEmeralds asEnchantBookForEmeralds();  
      
    IEnchantedItemForEmeralds asEnchantedItemForEmeralds();  
      
    IItemsForEmeralds asItemsForEmeralds();  
      
    ISuspiciousStewForEmerald asSuspiciousStewForEmerald();  
      
    ITreasureMapForEmeralds asTreasureMapForEmeralds();  
}
```

`ITreasureMapForEmeralds` as an example would implement ways to change the destination or the map type.

### Image to see which fields belongs kinda together
image


MaxNeedsSnacks:
Right, first off, thanks for contributing, we've already talked about this a bit in DMs and on the Discord server, but basically, I feel like we should address this slightly differently:

- We should stick by the current "registry" type approach rather than using events, and especially take care to avoid `Consumer<X>` style events as those can lead to unnecessary runtime overhead.
- I'm not entirely sure whether we need this gigantic abstraction to `IItemListing` or any of those subclasses either, since vanilla's `MerchantOffer` (and children) already gives us everything required for this, doesn't it?
*Side note: The `I` prefix for interfaces is a practice we haven't adopted for this project, so we would probably want to rename these anyways to maintain consistency here*
- As a general guideline, I think we should stick to something close to our existing [`BiomeModifications`](https://github.com/architectury/architectury-api/blob/1.17/common/src/main/java/dev/architectury/registry/level/biome/BiomeModifications.java) API, maybe with the additional constraint of being able to reload this trade list at runtime (so either on server start or datapack reload)

MaxNeedsSnacks:
As an example suggestion, what if the general API structure looked like this?
```java
interface TradeContext { /* util interface based on a combination of a villager type, level and merchant offer */ }

// [...]

// remove and modify also exist
void addTrades(Predicate<TradeContext> predicate, BiConsumer<TradeContext, MerchantOffer> modifier)
```

LLytho:
Which kind of using `MerchantOffer` you think about? 

This:

```java
    @Mixin(targets = "net.minecraft.world.entity.npc.VillagerTrades$ItemsForEmeralds")
    static class ItemsForEmeraldsMixin implements IItemsForEmeralds {
    
        /**
         * Example of storing
         */
        MerchantOfferData data; 
        
        @Inject(method = "getOffer", at = @At("RETURN"), cancellable = true)
        public void onGetOffer(Entity entity, Random random, CallbackInfoReturnable<MerchantOffer> cir) {
            MerchantOffer offer = cir.getReturnValue();
            data.fill(offer);
        }
    }
```
So on `modify` we could only provide `MerchantOfferData`.  (from ur comment above, maybe `TradeContext`)
Q: Should a modder know in which trade type he is or do he only should have offer data?

Or more the way of always call an event when `getOffer()` is used in `AbstractVillager` & `WanderingTrader`

LLytho:
After discord discuss: 


- Create `MerchantOfferAccess` to provide an interface to change fields in `MerchantOffer`
- Inject `AbstractVillager::addOffersFromItemListings`, `WanderingTrader::addOffersFromItemListings` and `WanderingTrader::updateTrades` at the point where `itemListing.getOffer` creates an offer. 
- Call `handleFoo(Context context)` // name tbd 
- Context should provide: ItemListing, ItemListingType,, Entity, Random, @Nullable MerchantOfferAccess, level, VillagerProfession

- add `TradeRegistry::registerModify(VillagerProfession, level, Consumer<Context> consumer)` 
- add `TradeRegistry::registerRemove(VillagerProfession, level, Predicate<Context> predicate)` 

What else?

MaxNeedsSnacks:
Apart from my above comments, maybe since `Context` contains the profession and level anyways, the general structure could be simplified to
```java
addTrades(Predicate<Context> filter, Consumer<Context> callback)
```
so combining that with the above suggestion we'd have something like
```java
// called in that order
// context should be immutable during evaluation, mutable during callback, so maybe a subclass like below would be nice?
addTrades(Predicate<Context> filter, Consumer<Context.Mutable> callback)
removeTrades(Predicate<Context> filter, Consumer<Context.Mutable> callback)
modifyTrades(Predicate<Context> filter, Consumer<Context.Mutable> callback)
```

lythowastaken:
`removeTrades(Predicate<Context> filter, Consumer<Context.Mutable> callback)` 
what would be `callback` here? Shouldn't the filter be the decider? And I thought because you were worry about `Consumer` overhead, doing `(profession, level, ...)` would be healthier.  And with just profession and level you can hold a map with the consumer and call `handleFoo` only for needed consumer.  Otherwise you run `handleFoo` over everything. Filtering can still happen through the consumer when `handleFoo` is called. 

```java
modifyTrades(FARMER, 2, (contex) -> {
  if(context.trade.getEmeraldCost() > 2) { return; }
  
  // handle
});
```

MaxNeedsSnacks:
> what would be callback here? Shouldn't the filter be the decider?

I think the naming may be a bit misleading here haha
The method name in this case doesn't literally mean "remove these trades", but rather, "do this with the trades during the `REMOVAL` phase of the event". The names here are really only suggestions in this case (you could also just use HIGH, NORMAL, LOW, to signify the "priorities" of these phases) and are mostly based on the fabric event phases for BiomeModifications (which we basically replicate for architectury as well)

MaxNeedsSnacks:
Addressed the misunderstanding on Discord, sorry, don't mean to bikeshed haha

LLytho:
While looking onto the code: 
image

So probably the solution would be create the full set of random `int`s but just use 2 as long nothing is skipped or removed. Not really sure if this is a good solution.

shedaniel:
Is this ready for review? Please also properly rebase the PR onto the main branch, this is currently showing unrelated commits.

LLytho:
> Is this ready for review? Please also properly rebase the PR onto the main branch, this is currently showing unrelated commits.

Will mark it as ready. Yeah I fucked up the commit. I merged `origin/1.16` instead of rebase `upstream/1.16`.

LLytho:
You cannot see this. I fucked up. Look here:
https://github.com/architectury/architectury-api/pull/122

PR #122, opened 2021-07-28
Feature/villager trades

LLytho:
Register modify or removing for trade offers:
`TradeRegistry.registerVillagerOfferModify(Consumer<VillagerTradeOfferContext> consumer)`
`TradeRegistry.registerVillagerOfferRemoving(Predicate<VillagerTradeOfferContext> predicate)`
`TradeRegistry.registerWanderingTraderOfferModify(Consumer<WanderingTraderOfferContext> consumer)`
`TradeRegistry.registerWanderingTraderOfferRemoving(Predicate<WanderingTraderOfferContext> predicate)`

Override the max possible offers for a villager or the wandering trader. As an example wandering trader has over 50 possible trades. Vanilla just select 5 per default.
`TradeRegistry.overrideVillagerMaxOffers(VillagerProfession profession, int level, int maxOffers)`
`TradeRegistry.overrideWanderingTraderMaxOffers(int maxOffers)`

### DIFF
--- a/common/src/main/java/me/shedaniel/architectury/mixin/AbstractVillagerMixin.java
+++ b/common/src/main/java/me/shedaniel/architectury/mixin/AbstractVillagerMixin.java
@@ -1,0 +1,87 @@
+/*
+ * This file is part of architectury.
+ * Copyright (C) 2020, 2021 architectury
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+package me.shedaniel.architectury.mixin;
+
+import me.shedaniel.architectury.registry.trade.OfferMixingContext;
+import net.minecraft.world.entity.Entity;
+import net.minecraft.world.entity.EntityType;
+import net.minecraft.world.entity.npc.AbstractVillager;
+import net.minecraft.world.entity.npc.VillagerTrades;
+import net.minecraft.world.item.trading.MerchantOffer;
+import net.minecraft.world.item.trading.MerchantOffers;
+import net.minecraft.world.level.Level;
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.Unique;
+import org.spongepowered.asm.mixin.injection.At;
+import org.spongepowered.asm.mixin.injection.ModifyVariable;
+import org.spongepowered.asm.mixin.injection.Redirect;
+
+import java.util.Iterator;
+import java.util.Set;
+
+
+@Mixin(AbstractVillager.class)
+public abstract class AbstractVillagerMixin extends Entity {
+    public AbstractVillagerMixin(EntityType<?> entityType, Level level) {
+        super(entityType, level);
+    }
+    
+    @Unique
+    private final ThreadLocal<OfferMixingContext> offerContext = new ThreadLocal<>();
+    
+    @Redirect(
+            method = "addOffersFromItemListings(Lnet/minecraft/world/item/trading/MerchantOffers;[Lnet/minecraft/world/entity/npc/VillagerTrades$ItemListing;I)V",
+            at = @At(value = "INVOKE", target = "Ljava/util/Set;iterator()Ljava/util/Iterator;")
+    )
+    public Iterator<Integer> overrideIterator(Set<Integer> set, MerchantOffers offers, VillagerTrades.ItemListing[] itemListings, int maxOffers) {
+        OfferMixingContext context = new OfferMixingContext(overrideMaxOffersIfRegistered(maxOffers), itemListings, random);
+        offerContext.set(context);
+        return context.getIterator();
+    }
+    
+    @ModifyVariable(
+            method = "addOffersFromItemListings(Lnet/minecraft/world/item/trading/MerchantOffers;[Lnet/minecraft/world/entity/npc/VillagerTrades$ItemListing;I)V",
+            at = @At(value = "STORE"), // TODO ask someone why `INVOKE_ASSIGN` triggers `merchantOffers.add(merchantOffer)` too
+            ordinal = 0
+    )
+    public MerchantOffer handleOffer(MerchantOffer offer) {
+        OfferMixingContext context = offerContext.get();
+        
+        if (offer == null || context.getMaxOffers() == 0) {
+            context.skipIteratorIfMaxOffersReached();
+            return null;
+        }
+        
+        MerchantOffer handledOffer = handleOfferImpl(offer);
+        if (handledOffer != null) {
+            context.skipIteratorIfMaxOffersReached();
+        }
+        
+        return handledOffer;
+    }
+    
+    public MerchantOffer handleOfferImpl(MerchantOffer offer) {
+        return offer;
+    }
+    
+    public int overrideMaxOffersIfRegistered(int currentMaxOffers) {
+        return currentMaxOffers;
+    }
+}
--- a/common/src/main/java/me/shedaniel/architectury/mixin/AbstractVillagerMixin.java
+++ b/common/src/main/java/me/shedaniel/architectury/mixin/AbstractVillagerMixin.java
@@ -20,8 +20,8 @@
 package me.shedaniel.architectury.mixin;
 
 import com.google.common.base.MoreObjects;
-import me.shedaniel.architectury.registry.trade.impl.OfferMixingContext;
 import me.shedaniel.architectury.registry.trade.TradeRegistry;
+import me.shedaniel.architectury.registry.trade.impl.OfferMixingContext;
 import net.minecraft.world.entity.Entity;
 import net.minecraft.world.entity.EntityType;
 import net.minecraft.world.entity.npc.AbstractVillager;
@@ -49,7 +49,6 @@ import java.util.Set;
  * <p>
  * To solve this we override the iterator with our own iterator which iterate through all indexes.
  * As soon {@link OfferMixingContext#maxOffers} offers are created we skip the remaining elements in the iterator {@link OfferMixingContext#skipIteratorIfMaxOffersReached()}.
- *
  */
 @Mixin(AbstractVillager.class)
 public abstract class AbstractVillagerMixin extends Entity {
@@ -60,7 +59,7 @@ public abstract class AbstractVillagerMixin extends Entity {
     @Unique
     private final ThreadLocal<OfferMixingContext> offerContext = new ThreadLocal<>();
     
- 
+    
     @Redirect(
             method = "addOffersFromItemListings(Lnet/minecraft/world/item/trading/MerchantOffers;[Lnet/minecraft/world/entity/npc/VillagerTrades$ItemListing;I)V",
             at = @At(value = "INVOKE", target = "Ljava/util/Set;iterator()Ljava/util/Iterator;")
--- a/common/src/main/java/me/shedaniel/architectury/mixin/AbstractVillagerMixin.java
+++ b/common/src/main/java/me/shedaniel/architectury/mixin/AbstractVillagerMixin.java
@@ -20,7 +20,7 @@
 package me.shedaniel.architectury.mixin;
 
 import com.google.common.base.MoreObjects;
-import me.shedaniel.architectury.registry.trade.interal.OfferMixingContext;
+import me.shedaniel.architectury.registry.trade.impl.OfferMixingContext;
 import net.minecraft.world.entity.Entity;
 import net.minecraft.world.entity.EntityType;
 import net.minecraft.world.entity.npc.AbstractVillager;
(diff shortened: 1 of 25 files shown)


common/src/main/java/me/shedaniel/architectury/registry/trade/TradeRegistry.java
```diff
@@ -59,4 +144,38 @@ public static void registerTradeForWanderingTrader(boolean rare, VillagerTrades.
...
+    }
+    
+    /**
+     * @return Max offers for the wandering trader. Returning null means no override exists
+     */
+    public static Integer getWanderingTraderMaxOffers() {
+        return WANDERING_TRADER_MAX_OFFER_OVERRIDE;
+    }
+    
+    public static boolean invokeVillagerOfferRemoving(VillagerTradeOfferContext ctx) {
```
MaxNeedsSnacks: These should not be publicly accessible to users imo, or at the very least not be included in the same class as the user-facing API. Consider moving this (as well as perhaps the maps themselves) to another class marked as internal.

common/src/main/java/me/shedaniel/architectury/registry/trade/TradeRegistry.java
```diff
@@ -47,6 +59,79 @@ private static void registerVillagerTrade0(VillagerProfession profession, int le
...
+        Objects.requireNonNull(predicate);
+        VILLAGER_REMOVE_HANDLERS.add(predicate);
+    }
+    
+    /**
+     * Register a consumer which provide {@link WanderingTraderOfferContext} to modify the given offer from the wandering trader.
+     * The consumer gets called when {@link net.minecraft.world.entity.npc.WanderingTrader} generates their offer list.
+     * @param consumer The consumer to handle modification for the given offer context.
+     */
+    public static void registerWanderingTraderOfferModify(Consumer<WanderingTraderOfferContext> consumer) {
```
MaxNeedsSnacks:
```suggestion
    public static void modifyWanderingTraderOffers(Consumer<WanderingTraderOfferContext> callback) {
```

common/src/main/java/me/shedaniel/architectury/registry/trade/TradeRegistry.java
```diff
@@ -47,6 +59,79 @@ private static void registerVillagerTrade0(VillagerProfession profession, int le
...
+        Objects.requireNonNull(consumer);
+        VILLAGER_MODIFY_HANDLERS.add(consumer);
+    }
+    
+    /**
+     * Register a predicate which provide {@link VillagerTradeOfferContext} to test the given offer from a villager.
+     * The predicate gets called when {@link net.minecraft.world.entity.npc.Villager} generates their offer list.
+     * @param predicate The predicate to test if an offer should be removed. Returning true means the offer will be removed.
+     */
+    public static void registerVillagerOfferRemoving(Predicate<VillagerTradeOfferContext> predicate) {
```
MaxNeedsSnacks:
```suggestion
    public static void removeVillagerOffers(Predicate<VillagerTradeOfferContext> filter) {
```

common/src/main/java/me/shedaniel/architectury/registry/trade/TradeRegistry.java
```diff
@@ -47,6 +59,79 @@ private static void registerVillagerTrade0(VillagerProfession profession, int le
...
+        return VILLAGER_MAX_OFFER_OVERRIDES.get(profession).get(level);
+    }
+    
+    
+    /**
+     * Register a consumer which provide {@link VillagerTradeOfferContext} to modify the given offer from a villager.
+     * The consumer gets called when {@link net.minecraft.world.entity.npc.Villager} generates their offer list.
+     * @param consumer The consumer to handle modification for the given offer context.
+     */
+    public static void registerVillagerOfferModify(Consumer<VillagerTradeOfferContext> consumer) {
```
MaxNeedsSnacks:
```suggestion
    public static void modifyVillagerOffers(Consumer<VillagerTradeOfferContext> callback) {
```

common/src/main/java/me/shedaniel/architectury/registry/trade/TradeRegistry.java
```diff
@@ -47,6 +59,79 @@ private static void registerVillagerTrade0(VillagerProfession profession, int le
         throw new AssertionError();
     }
     
+    /**
+     * Override the max possible offers a villager can have by its profession and level.
+     * @param profession The Profession of the villager.
+     * @param level      The level of the villager. Vanilla range is 1 to 5, however mods may extend that upper limit further.
+     * @param maxOffers  Max possible offers a villager can have.
+     */
+    public static void overrideVillagerMaxOffers(VillagerProfession profession, int level, int maxOffers) {
```
MaxNeedsSnacks:
```suggestion
    public static void setVillagerMaxOffers(VillagerProfession profession, int level, int maxOffers) {
```

common/src/main/java/me/shedaniel/architectury/registry/trade/TradeRegistry.java
```diff
@@ -59,4 +144,38 @@ public static void registerTradeForWanderingTrader(boolean rare, VillagerTrades.
         throw new AssertionError();
     }
     
+    /**
+     * Override the max possible offers the wandering trader can have. This does not affect the rare trade.
+     * @param maxOffers  Max possible offers a villager can have.
+     */
+    public static void overrideWanderingTraderMaxOffers(int maxOffers) {
```
MaxNeedsSnacks:
```suggestion
    public static void setWanderingTraderMaxOffers(int maxOffers) {
```

common/src/main/java/me/shedaniel/architectury/registry/trade/TradeRegistry.java
```diff
@@ -47,6 +59,79 @@ private static void registerVillagerTrade0(VillagerProfession profession, int le
...
+        Objects.requireNonNull(consumer);
+        WANDERING_TRADER_MODIFY_HANDLERS.add(consumer);
+    }
+    
+    /**
+     * Register a predicate which provide {@link WanderingTraderOfferContext} to test the given offer from the wandering trader.
+     * The predicate gets called when {@link net.minecraft.world.entity.npc.WanderingTrader} generates their offer list.
+     * @param predicate The predicate to test if an offer should be removed. Returning true means the offer will be removed.
+     */
+    public static void registerWanderingTraderOfferRemoving(Predicate<WanderingTraderOfferContext> predicate) {
```
MaxNeedsSnacks:
```suggestion
    public static void removeWanderingTraderOffers(Predicate<WanderingTraderOfferContext> filter) {
```

MaxNeedsSnacks (changes requested):
Only managed to have a cursory glance at these as I'm a bit strapped for time, so this isn't a full review yet, but I think these things may need addressing

MaxNeedsSnacks:
Sorry for the long wait, I'll be checking this out soon ^^
(That said @shedaniel also seems to have taken interest in it seeing as he self-assigned this, so I might also just wait for his review here)

common/src/main/java/me/shedaniel/architectury/registry/trade/TradeRegistry.java
```diff
@@ -23,7 +23,19 @@
...
+import java.util.function.Predicate;
+
 public class TradeRegistry {
+    private static final List<Consumer<VillagerTradeOfferContext>> VILLAGER_MODIFY_HANDLERS = new ArrayList<>();
+    private static final List<Predicate<VillagerTradeOfferContext>> VILLAGER_REMOVE_HANDLERS = new ArrayList<>();
+    private static final List<Consumer<WanderingTraderOfferContext>> WANDERING_TRADER_MODIFY_HANDLERS = new ArrayList<>();
+    private static final List<Predicate<WanderingTraderOfferContext>> WANDERING_TRADER_REMOVE_HANDLERS = new ArrayList<>();
+    
+    private static final Map<VillagerProfession, Map<Integer, Integer>> VILLAGER_MAX_OFFER_OVERRIDES = new HashMap<>();
+    private static Integer WANDERING_TRADER_MAX_OFFER_OVERRIDE = null;
```
shedaniel: wrong case

common/src/main/java/me/shedaniel/architectury/registry/trade/TradeRegistry.java
```diff
@@ -59,4 +144,38 @@ public static void registerTradeForWanderingTrader(boolean rare, VillagerTrades.
...
+            throw new IllegalArgumentException("Wandering trader's max offers has to be at least 0!");
+        }
+        
+        WANDERING_TRADER_MAX_OFFER_OVERRIDE = maxOffers;
+    }
+    
+    /**
+     * @return Max offers for the wandering trader. Returning null means no override exists
+     */
+    public static Integer getWanderingTraderMaxOffers() {
```
shedaniel: getWanderingTraderMaxOffersOverride or something along the lines of that but shorter

also @Nullable

common/src/main/java/me/shedaniel/architectury/mixin/WanderingTraderMixin.java
```diff
@@ -0,0 +1,94 @@
...
+            return null;
+        }
+        
+        TradeRegistry.invokeWanderingTraderOfferModify(context);
+        return offer;
+    }
+    
+    @Override
+    public int overrideMaxOffersIfRegistered(int currentMaxOffers) {
+        Integer newMaxOffers = TradeRegistry.getWanderingTraderMaxOffers();
```
shedaniel: could use MoreObjects.firstNonNull

common/src/main/java/me/shedaniel/architectury/mixin/AbstractVillagerMixin.java
```diff
@@ -0,0 +1,87 @@
...
+        
+        MerchantOffer handledOffer = handleOfferImpl(offer);
+        if (handledOffer != null) {
+            context.skipIteratorIfMaxOffersReached();
+        }
+        
+        return handledOffer;
+    }
+    
+    public MerchantOffer handleOfferImpl(MerchantOffer offer) {
```
shedaniel: this needs to have architectury$ prefixed

common/src/main/java/me/shedaniel/architectury/registry/trade/MerchantOfferAccess.java
```diff
@@ -0,0 +1,79 @@
...
+import net.minecraft.world.item.trading.MerchantOffer;
+
+public class MerchantOfferAccess {
+    private final MerchantOffer offer;
+    
+    public MerchantOfferAccess(MerchantOffer offer) {
+        this.offer = offer;
+    }
+    
+    public ItemStack getPrimary() {
```
shedaniel: these should be getPrimaryCost or exactly same as the mojmap ones

common/src/main/java/me/shedaniel/architectury/registry/trade/MerchantOfferAccess.java
```diff
@@ -0,0 +1,79 @@
...
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+package me.shedaniel.architectury.registry.trade;
+
+import net.minecraft.world.item.ItemStack;
+import net.minecraft.world.item.trading.MerchantOffer;
+
+public class MerchantOfferAccess {
+    private final MerchantOffer offer;
```
shedaniel, line 26: might want a getter for this field

shedaniel (commented):
Pretty good, some minor issues, will test this locally later

common/src/main/java/me/shedaniel/architectury/registry/trade/MerchantOfferAccess.java
LLytho:
```java
public class MerchantOffer {
   public ItemStack baseCostA;
   public ItemStack costB;
   public ItemStack result;
   // ...
}
```

So better `getCostA` and `getCostB` or maybe `getPrimaryPrice` like in `SimpleTrade`?

common/src/main/java/me/shedaniel/architectury/registry/trade/TradeRegistry.java
LLytho: Without just shorten `WanderingTrader` into `WT` what would be a shorter version for it? I don't think in general longer method names are bad.

 added @nullable for `getVillagerMaxOffers` too

common/src/main/java/me/shedaniel/architectury/registry/trade/TradeRegistry.java
LLytho: just for `WANDERING_TRADER_MAX_OFFER_OVERRIDE` here right?

shedaniel: yeah because it is not static final

common/src/main/java/me/shedaniel/architectury/registry/trade/TradeRegistry.java
shedaniel: tbh considering this again maybe it is best to not expose this method in the "api", using WT is kinda meh and I don't feel like that's where we want to simplify at all, would 100% want the full WanderingTrader instead

common/src/main/java/me/shedaniel/architectury/registry/trade/MerchantOfferAccess.java
shedaniel: I would use cost instead of price here, but basically yes

common/src/main/java/me/shedaniel/architectury/registry/trade/TradeRegistry.java
LLytho: With this. I would move everything into a `trading` package and just leave `MerchantOfferAccess` and the `TradeRegistry` left in the current package? And then move the invoke stuff into his own class. Something against this?

MaxNeedsSnacks: I'd rather have the package be called `internal` or `impl` as we **almost definitely** don't want users to call this themselves, so highlighting this with a different package and `@ApiStatus.Internal` might help with that

MaxNeedsSnacks (approved):

PR #128, opened 2021-08-11
Add tool interaction hooks

Leo40Git:
This PR adds 3 hooks to add "right-click with tool" interactions:
- Stripping (w/axe) 
- Flattening (w/shovel)
- Tilling (w/hoe)

### DIFF
--- a/common/src/main/java/dev/architectury/hooks/item/tool/AxeItemHooks.java
+++ b/common/src/main/java/dev/architectury/hooks/item/tool/AxeItemHooks.java
@@ -3,6 +3,7 @@ package dev.architectury.hooks.item.tool;
 import com.google.common.collect.ImmutableMap;
 import net.minecraft.world.item.AxeItem;
 import net.minecraft.world.level.block.Block;
+import net.minecraft.world.level.block.RotatedPillarBlock;
 
 import java.util.HashMap;
 
@@ -11,6 +12,10 @@ public final class AxeItemHooks {
     }
 
     public static void addStrippingRecipe(Block src, Block dst) {
+        if (!src.defaultBlockState().hasProperty(RotatedPillarBlock.AXIS))
+            throw new IllegalArgumentException("Source block is missing required 'AXIS' property!");
+        if (!dst.defaultBlockState().hasProperty(RotatedPillarBlock.AXIS))
+            throw new IllegalArgumentException("Destination block is missing required 'AXIS' property!");
         if (AxeItem.STRIPPABLES instanceof ImmutableMap) {
             AxeItem.STRIPPABLES = new HashMap<>(AxeItem.STRIPPABLES);
         }
--- a/common/src/main/java/dev/architectury/hooks/item/tool/AxeItemHooks.java
+++ b/common/src/main/java/dev/architectury/hooks/item/tool/AxeItemHooks.java
@@ -1,0 +1,19 @@
+package dev.architectury.hooks.item.tool;
+
+import com.google.common.collect.ImmutableMap;
+import net.minecraft.world.item.AxeItem;
+import net.minecraft.world.level.block.Block;
+
+import java.util.HashMap;
+
+public final class AxeItemHooks {
+    private AxeItemHooks() {
+    }
+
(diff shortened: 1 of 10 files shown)


common/src/main/java/dev/architectury/hooks/item/tool/AxeItemHooks.java
```diff
@@ -0,0 +1,19 @@
...
+
+public final class AxeItemHooks {
+    private AxeItemHooks() {
+    }
+
+    public static void addStrippingRecipe(Block src, Block dst) {
+        if (AxeItem.STRIPPABLES instanceof ImmutableMap) {
+            AxeItem.STRIPPABLES = new HashMap<>(AxeItem.STRIPPABLES);
+        }
+        AxeItem.STRIPPABLES.put(src, dst);
```
Juuxel: Does the axe item code still assume the block has `AXIS` (ie is a `PillarBlock`)?

Leo40Git: Yes, it assumes both the source block _and_ the destination block have the `AXIS` property.

Juuxel: It'd probably be good to add checks here already so that it crashes already at registration time instead of when the recipe is used.

common/src/main/java/dev/architectury/hooks/item/tool/HoeItemHooks.java
```diff
@@ -0,0 +1,23 @@
...
+
+import java.util.HashMap;
+import java.util.function.Consumer;
+import java.util.function.Predicate;
+
+public class HoeItemHooks {
+    private HoeItemHooks() {
+    }
+
+    public static void addTillingRecipe(Block src, Predicate<UseOnContext> predicate, Consumer<UseOnContext> action) {
```
Juuxel: Javadoc is especially needed here - what does the predicate do? What does the action do?

Juuxel (changes requested):
The registration methods should be documented with javadoc (at least the hoe one, see my comments there). The API itself is nice and simple!

I'd also prefer `input` and `result` to `src` and `dst`, but that's just a personal style opinion so no need to change it either IMO.

common/src/main/java/dev/architectury/hooks/item/tool/HoeItemHooks.java
Leo40Git: TL;DR is: Game invokes `predicate.test(ctx)`, if that returns true the game first invokes `action.accept(ctx)` and then damages the hoe by 1.

Leo40Git:
Javadoc is here. I don't know how good it is, but it _is_ there.

Juuxel (approved):

common/src/main/java/dev/architectury/hooks/item/tool/AxeItemHooks.java
```diff
@@ -0,0 +1,37 @@
...
+     * {@link net.minecraft.world.level.block.state.properties.BlockStateProperties#AXIS AXIS} property,
+     * and that the value of this property will be copied from the input block to the result block when the recipe
+     * is performed.
+     *
+     * @param input input block
+     * @param result result block
+     * @throws IllegalArgumentException if the input or result blocks do not have the
+     * {@link net.minecraft.world.level.block.state.properties.BlockStateProperties#AXIS AXIS} property.
+     */
+    public static void addStrippingRecipe(Block input, Block result) {
```
shedaniel: Should this be renamed to addStrippable? No names here relates to recipes

testmod-common/src/main/java/dev/architectury/test/TestMod.java
```diff
@@ -52,6 +53,7 @@ public static void initialize() {
         TestTrades.init();
         TestParticles.initialize();
         TestModNet.initialize();
+        TestRecipes.init();
```
shedaniel: Should this be TestBlockInteractions

shedaniel (approved):

PR #144, opened 2021-11-28
Remove hacks around RegistryEntry

shedaniel:
(Courtesy of OrionOnline)

Extending `ForgeRegistryEntry` is not required to implement `IForgeRegistryEntry`, previous concerns for explicit casts for `ForgeRegistryEntry` are not applicable to registries whose elements are interfaces.
This PR simplifies our setup for these entries, and adds methods that would extend methods of `IForgeRegistryEntry`.

This PR should not impact mods negatively, but it should be looked into if there are any architectury mods casting entries to `ForgeRegistryEntry` directly.


MaxNeedsSnacks:
Uh, what happened with that AW on the Java 16 build? :p

Juuxel:
The AW is invalid (probably references a removed field), so Loom's new-ish `validateAccessWidener` catches it during the build.

MaxNeedsSnacks (approved):

Juuxel (approved):

Colin-J-Wood:
This fix is causing a crash on mods that are expecting the original implementation.  we are unsure if it is the api's fault or the mod's.  I have mentioned the mod in question below.

shedaniel:
Tracking #150, I will look into this later

PR #147, opened 2021-12-06
Transfer & Lookup

### LINKED ISSUE #15: Lookup API
API to lookup components for objects, should be extendable to allow hooking up to capabilities and cardinal components.

Injecting custom components should be deferred and delegated to runtime, ASM may be used to generate the capabilities getters.

It would be best of our interest to wrap around the APIs without delegation to avoid creating additional objects.

shedaniel:
This PR adds abstractions on top of Fabric's Lookup+Transfer API and Forge's Item/FluidHandler API.
Support for abstractions for energy can be considered in the future as it shouldn't be too hard to do them.

# Usage
## Lookup
`ItemTransfer.BLOCK` and `FluidTransfer.BLOCK` provides access to Fabric's Lookup API and Forge's capabilities.
`BlockLookup` provides access to other blocks with the Level, BlockPos, or the conditional Context.

#### Fabric
The context, `Direction`, is not-nullable, and must be provided in order to query.

#### Forge
The lookup only functions if the targeted block has a BlockEntity, since capabilities are only attached on the BlockEntity level, this is different from Fabric, where it can be available. Examples of these are cauldrons, Fabric provide implementations for them, while it is impossible to support them on Forge.

### Examples
```java
TransferHandler<FluidStack> above = FluidTransfer.BLOCK.get(getLevel(), getBlockPos().above(), Direction.DOWN);

// Handlers returned by the lookup may be null
if (above != null) {
    // Do things...
}
```

[insert more shit here]


vini2003:
amazing do thanks

KP2048:
How close is this to being merged? i have my own energy api that uses architectury, but this one is better because it also does item and fluid.

MaxNeedsSnacks:
IIRC Dan has said he wants to move this to an Architectury Extras repo as to not clog up the main API, so we'd have to work on getting that one up and running first; other than that, I'm not sure what (if anything) is missing here still ^^

PR #148, opened 2021-12-08
Add LifecycleEvent.SETUP for manipulating things after normal initial…

shedaniel:
…ization

Should be useful for setting up (properties?) for when registries _should_ be done.
I know fabric doesn't have a hard limit on where you can register things, but this should be useful enough for most cases.

### DIFF
--- a/common/src/main/java/dev/architectury/event/events/common/LifecycleEvent.java
+++ b/common/src/main/java/dev/architectury/event/events/common/LifecycleEvent.java
@@ -21,6 +21,7 @@ package dev.architectury.event.events.common;
 
 import dev.architectury.event.Event;
 import dev.architectury.event.EventFactory;
+import dev.architectury.event.events.client.ClientLifecycleEvent;
 import net.minecraft.server.MinecraftServer;
 import net.minecraft.server.level.ServerLevel;
 import net.minecraft.world.level.Level;
@@ -88,6 +89,14 @@ public interface LifecycleEvent {
      * @see ServerLevelState#act(Level)
      */
     Event<ServerLevelState> SERVER_LEVEL_SAVE = EventFactory.createLoop();
+    /**
+     * Invoked once common setup has begun.
+     * <p> This happens during {@code FMLCommonSetupEvent} on Forge,
+     * or when Architectury API's entrypoint initialises on Fabric.
+     * <p>
+     * Registries should be initialised here.
+     */
+    Event<Runnable> SETUP = EventFactory.createLoop();
     
     interface InstanceState<T> {
         /**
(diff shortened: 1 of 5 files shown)


common/src/main/java/dev/architectury/event/events/common/LifecycleEvent.java
```diff
@@ -88,6 +89,14 @@
      * @see ServerLevelState#act(Level)
      */
     Event<ServerLevelState> SERVER_LEVEL_SAVE = EventFactory.createLoop();
+    /**
+     * Invoked once common setup has begun.
+     * <p> This happens during {@code FMLCommonSetupEvent} on Forge,
+     * or when Architectury API's entrypoint initialises on Fabric.
```
BasiqueEvangelist:
```suggestion
     * or when Architectury API's client/server entrypoint initialises on Fabric.
```

common/src/main/java/dev/architectury/event/events/common/LifecycleEvent.java
```diff
@@ -88,6 +89,14 @@
...
+     * <p>
+     * Registries should be initialised here.
```
MaxNeedsSnacks: To a new user, "Registries should be initialised here." implies that **they** should initialise their registries here, which would be incorrect. This should probably be rephrased to "Registries should have been initialised by this point."

common/src/main/java/dev/architectury/event/events/common/LifecycleEvent.java
shedaniel: I don't think this is important and I think this may be even more confusing, fabric simply invokes the client and server entrypoints after the main one, so the main reason why we put our hook there is to be invoked after all of the main ones.

BasiqueEvangelist: Yes, but doesn't *this* imply that it's when Architectury API's `ModInitializer` is invoked?

shedaniel: I mean this doesn’t say main entrypoint

MaxNeedsSnacks: I feel like people might assume "entrypoint" would just mean "main" entrypoint, so Basique does have a point here, though I'm not sure whether "client/server entrypoint" is better terminology here

shedaniel: I am accepting Basique's suggestion here, since I can't think of a better javadoc.

MaxNeedsSnacks (approved):

PR #154, opened 2021-12-18
@Redirect workaround for closing #152

### LINKED ISSUE #152: Conflict with Mouse Tweaks due to DelegateScreen wrapping
Hi, I'm the developer of [Mouse Tweaks](https://github.com/YaLTeR/MouseTweaks). I recently added a Fabric version and just got a [bug report](https://github.com/YaLTeR/MouseTweaks/issues/62) that Mouse Tweaks doesn't work when Architectury is loaded.

The way Mouse Tweaks currently works is it [subscribes](https://github.com/YaLTeR/MouseTweaks/blob/aaa09e47fd171ff5086862a0abda1ef8c57af515/src/main/java/yalter/mousetweaks/fabric/MouseTweaksFabric.java#L15) to every `Screen`'s mouse events (with [its own mixin](https://github.com/YaLTeR/MouseTweaks/blob/aaa09e47fd171ff5086862a0abda1ef8c57af515/src/main/java/yalter/mousetweaks/fabric/mixin/MixinMouseHandler.java#L21) for the dragging as there's no hook in Fabric for this), and [checks](https://github.com/YaLTeR/MouseTweaks/blob/aaa09e47fd171ff5086862a0abda1ef8c57af515/src/main/java/yalter/mousetweaks/Main.java#L219) in the callbacks if it can handle the `screen` that it gets. In particular, the check [succeeds](https://github.com/YaLTeR/MouseTweaks/blob/aaa09e47fd171ff5086862a0abda1ef8c57af515/src/main/java/yalter/mousetweaks/Main.java#L552) if `screen instanceof AbstractContainerScreen`.

Now this is what I'm seeing in the log upon holding and releasing my mouse with Architectury loaded:

```
[15:49:03] [Render thread/INFO]: [STDOUT]: [Mouse Tweaks] You have just opened a class_476.
[15:49:03] [Render thread/INFO]: [STDOUT]: [Mouse Tweaks] Handler: GuiContainerHandler; Mouse Tweaks is enabled; wheel tweak is enabled.
[15:49:04] [Render thread/INFO]: [STDOUT]: [Mouse Tweaks] You have just opened a DelegateScreen.
[15:49:04] [Render thread/INFO]: [STDOUT]: [Mouse Tweaks] No valid handler found; Mouse Tweaks is disabled.
[15:49:05] [Render thread/INFO]: [STDOUT]: [Mouse Tweaks] You have just opened a class_476.
[15:49:05] [Render thread/INFO]: [STDOUT]: [Mouse Tweaks] Handler: GuiContainerHandler; Mouse Tweaks is enabled; wheel tweak is enabled.
```

So Mouse Tweaks does recognize the screen initially, but for the duration of mouse hold and drag the events get delivered to a `DelegateScreen` which Mouse Tweaks does not recognize. I haven't checked yet but it might be specifically the drag handler that does this.

This seems like something Architectury does: https://github.com/architectury/architectury-api/blob/11ef83f9de1410f46830cb317fb25735924aa5bc/fabric/src/main/java/dev/architectury/mixin/fabric/client/MixinMouseHandler.java#L159

How should we fix this, should Architectury adjust the behavior somehow, or is there a workaround I can add to Mouse Tweaks?

Thanks.

shedaniel:
This reworks the mixin for implementing mouse drag events, this is more dangerous and incompatible compared to just wrapping the screen.

Whether there are other mods redirecting the same method should be looked into for this issue, the reason why I originally implemented this mixin with the `@ModifyVariable` is that I needed to get the return value of the vanilla `mouseDragged` method, we could potentially do raw asm here and add our transformation before the `pop`, but that would be way overkill.

```asm
 0 aload_1
 1 dload_2
 2 dload 4
 4 aload_0
 5 getfield #52 <net/minecraft/client/MouseHandler.activeButton : I>
 8 dload 6
10 dload 8
12 invokevirtual #482 <net/minecraft/client/gui/screens/Screen.mouseDragged : (DDIDD)Z>
15 pop <-- we need to inject before this
16 return
```

If I am missing some mixin magic features that could solve this issue without `@Redirect`, please do let me know.

### DIFF
--- a/fabric/src/main/java/dev/architectury/impl/fabric/ScreenInputDelegate.java
+++ b/fabric/src/main/java/dev/architectury/impl/fabric/ScreenInputDelegate.java
@@ -36,15 +36,6 @@ public interface ScreenInputDelegate {
         }
         
         @Override
-        public boolean mouseDragged(double d, double e, int i, double f, double g) {
-            if (ClientScreenInputEvent.MOUSE_DRAGGED_PRE.invoker().mouseDragged(Minecraft.getInstance(), parent, d, e, i, f, g).isPresent())
-                return true;
-            if (parent.mouseDragged(d, e, i, f, g))
-                return true;
-            return ClientScreenInputEvent.MOUSE_DRAGGED_POST.invoker().mouseDragged(Minecraft.getInstance(), parent, d, e, i, f, g).isPresent();
-        }
-        
-        @Override
         public boolean charTyped(char c, int i) {
             if (ClientScreenInputEvent.CHAR_TYPED_PRE.invoker().charTyped(Minecraft.getInstance(), parent, c, i).isPresent())
                 return true;
--- a/fabric/src/main/java/dev/architectury/mixin/fabric/client/MixinMouseHandler.java
+++ b/fabric/src/main/java/dev/architectury/mixin/fabric/client/MixinMouseHandler.java
@@ -165,6 +165,6 @@ public class MixinMouseHandler {
         if (screen.mouseDragged(mouseX, mouseY, button, deltaX, deltaY)) {
             return true;
         }
-        return ClientScreenInputEvent.MOUSE_DRAGGED_PRE.invoker().mouseDragged(Minecraft.getInstance(), screen, mouseX, mouseY, button, deltaX, deltaY).isPresent();
+        return ClientScreenInputEvent.MOUSE_DRAGGED_POST.invoker().mouseDragged(Minecraft.getInstance(), screen, mouseX, mouseY, button, deltaX, deltaY).isPresent();
     }
 }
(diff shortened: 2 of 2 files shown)


MaxNeedsSnacks (approved):
From what I can tell, there doesn't seem to be a better way to fix this using only mixins, so this is fine.

MaxNeedsSnacks:
cc @YaLTeR This would be a fix for the issue on our end, but since it's a *more* destructive Mixin than before, generally speaking, I'd love to see your input on this :p

YaLTeR:
Thanks for working on this!

First of all, this does fix the Mouse Tweaks issue.

As for whether this approach is good, I would consider two questions:
1. Are there any other mods doing anything with `onMouseDragged()`? Seems like a pretty specific place to deal with (but then again, here we are as two mods messing around in that exact area).
2. I think it would make sense to add `onMouseDragged()` events to Fabric rather than hardcoding the mixin in Mouse Tweaks. For Mouse Tweaks specifically, "allow" and "pre" events would be sufficient (which can be implemented with a simple HEAD mixin), however for "post" Fabric would have to do something along what this PR is doing. If Fabric were to do that, would that work for Architectury?

MaxNeedsSnacks:
The goal is definitely to get this into Fabric API itself at a later point, which would obviously then make this Mixin obsolete ^^

PR #159, opened 2021-12-23
Entity Model Layer Reg, and move to new package

shedaniel:
- Introduce `EntityModelLayerRegistry`, for adding to the ModelLayers map
- Move existing `EntityRendererRegistry` into the `client` package:
  OLD: `dev.architectury.registry.level.entity`
  NEW: `dev.architectury.registry.client.level.entity`


itsmeow (approved):
Looks good to me. I would use this extensively.

Juuxel (approved):

PR #189, opened 2022-01-31
Introduce item as an additional context

shedaniel:
Introduce an additional context for the item, it is specifically designed this way so other mods calling the normal tooltip render methods could attach the item context themselves (e.g. REI)


MaxNeedsSnacks (approved):

Juuxel:
How is this accessible for other mods outside arch api if the only way to get it is `@ApiStatus.Internal`?

shedaniel:
oh good point :concern:

Juuxel (approved):

PR #196, opened 2022-02-09
Fix particle factory registration for Forge to register at the correc…

jmilthedude:
…t time

- Added Architectury event to hook into Forge's ParticleFactoryRegistryEvent.
- Added resource test_particle json required for particles
- Removed deferred registration in Forge ParticleProviderRegistry impl, replaced with registry to proper forge event
- Removed EventBus and SubscribeEvent annotation which did not work (was not registered on the event bus properly)
- Fabric Impl untouched, works as it should (now that test_particle.json exists

)

### DIFF
--- a/common/src/main/java/dev/architectury/event/events/client/ClientParticleFactoryRegisterEvent.java
+++ b/common/src/main/java/dev/architectury/event/events/client/ClientParticleFactoryRegisterEvent.java
@@ -1,0 +1,17 @@
+package dev.architectury.event.events.client;
+
+import dev.architectury.event.Event;
+import dev.architectury.event.EventFactory;
+import net.fabricmc.api.EnvType;
+import net.fabricmc.api.Environment;
+
+@Environment(EnvType.CLIENT)
+public interface ClientParticleFactoryRegisterEvent {
+    /**
+     * Registers Particle Factory on the client.
+     * <p>
+     * <p>
+     * see Forge's {@code ParticleFactoryRegisterEvent}
+     */
+    Event<Runnable> REGISTER = EventFactory.createLoop();
+}
--- a/forge/src/main/java/dev/architectury/event/forge/EventHandlerImplClient.java
+++ b/forge/src/main/java/dev/architectury/event/forge/EventHandlerImplClient.java
@@ -328,5 +328,10 @@ public class EventHandlerImplClient {
         public static void event(RegisterShadersEvent event) {
             ClientReloadShadersEvent.EVENT.invoker().reload(event.getResourceManager(), event::registerShader);
         }
+        
+        @SubscribeEvent(priority = EventPriority.HIGH)
+        public static void event(ParticleFactoryRegisterEvent event) {
+            ClientParticleFactoryRegisterEvent.REGISTER.invoker().run();
+        }
     }
 }
--- a/forge/src/main/java/dev/architectury/registry/client/particle/forge/ParticleProviderRegistryImpl.java
+++ b/forge/src/main/java/dev/architectury/registry/client/particle/forge/ParticleProviderRegistryImpl.java
@@ -19,6 +19,7 @@
 
 package dev.architectury.registry.client.particle.forge;
 
+import dev.architectury.event.events.client.ClientParticleFactoryRegisterEvent;
 import dev.architectury.forge.ArchitecturyForge;
 import dev.architectury.registry.client.particle.ParticleProviderRegistry;
 import net.minecraft.client.Minecraft;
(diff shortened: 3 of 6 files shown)


common/src/main/java/dev/architectury/event/events/client/ClientParticleFactoryRegisterEvent.java
```diff
@@ -0,0 +1,17 @@
+package dev.architectury.event.events.client;
+
+import dev.architectury.event.Event;
+import dev.architectury.event.EventFactory;
+import net.fabricmc.api.EnvType;
+import net.fabricmc.api.Environment;
+
+@Environment(EnvType.CLIENT)
+public interface ClientParticleFactoryRegisterEvent {
```
shedaniel, line 9: I would rather have the event subscribe in the forge's ParticleProviderRegistryImpl fixed, we don't generally have these side-only events, only when it's necessary

jmilthedude: Fair point. I'll look into just fixing that rather than the route I went. Thanks for reviewing anyway.

jmilthedude: Hi again, here is why it doesn't seem like it will work. I fixed the event subscribe (all it needed was to point to the Mod event bus in the annotation because Forge bus is default). But the order of which the events are fired does not work. As can be seen in the log below, the particles initialize method is called, which is where the ParticleProviderRegistry.register function is added to the CLIENT_SETUP event. However, that event is fired later via FMLClientSetupEvent, after the ParticleFactoryRegistryEvent is fired. I do believe this makes the case for a new event, whether it is my suggestion or another one which handles an event prior to the ParticleFactoryRegistryEvent.

> TestParticles.initialize
onParticleFactoryRegister
[22:06:12] [Render thread/INFO] (Minecraft) Narrator library for x64 successfully loaded
[22:06:12] [Render thread/INFO] (Minecraft) Reloading ResourceManager: Default, Mod Resources
[22:06:12] [modloading-worker-0/WARN] (ForgeConfigSpec) Configuration file C:\Users\jmils\IdeaProjects\architectury-api\testmod-forge\run\config\forge-client.toml is not correct. Correcting
[22:06:12] [modloading-worker-0/WARN] (ForgeConfigSpec) Incorrect key client was corrected from null to its default, SimpleCommentedConfig:{}. 
[22:06:12] [modloading-worker-0/WARN] (ForgeConfigSpec) Incorrect key client.alwaysSetupTerrainOffThread was corrected from null to its default, false. 
[22:06:12] [modloading-worker-0/WARN] (ForgeConfigSpec) Incorrect key client.experimentalForgeLightPipelineEnabled was corrected from null to its default, false. 
[22:06:12] [modloading-worker-0/WARN] (ForgeConfigSpec) Incorrect key client.showLoadWarnings was corrected from null to its default, true. 
[22:06:12] [modloading-worker-0/WARN] (ForgeConfigSpec) Incorrect key client.useCombinedDepthStencilAttachment was corrected from null to its default, false. 
[22:06:12] [modloading-worker-0/WARN] (ForgeConfigSpec) Incorrect key client.forceSystemNanoTime was corrected from null to its default, false. 
[22:06:12] [modloading-worker-0/WARN] (ForgeConfigSpec) Configuration file C:\Users\jmils\IdeaProjects\architectury-api\testmod-forge\run\config\forge-common.toml is not correct. Correcting
[22:06:12] [modloading-worker-0/WARN] (ForgeConfigSpec) Incorrect key general was corrected from null to its default, SimpleCommentedConfig:{}. 
[22:06:12] [modloading-worker-0/WARN] (ForgeConfigSpec) Incorrect key general.defaultWorldType was corrected from null to its default, default. 
[22:06:12] [Render thread/INFO] (Architectury Test) Client started!
[22:06:12] [Worker-Main-23/ERROR] (Minecraft) Using missing texture, unable to load architectury_test:textures/mob_effect/test_effect.png : java.io.FileNotFoundException: architectury_test:textures/mob_effect/test_effect.png
FMLClientSetupEvent
ClientLifecycleEvent.CLIENT_SETUP.register

MaxNeedsSnacks:
@shedaniel @jmilthedude Can I get an update on the status of this? I'd obviously love to see this fixed in Architectury 1.18 so I'm not sure what kind of solution, if any, you have arrived at :p

jmilthedude:
I gave up on it. I was waiting for Architectury devs to come up with a solution that they liked better than what I came up with. The project I was working on wasn't very large anyway so I just wrote the Forge and Fabric versions separately without Architectury. I would love to see this fixed as well. It's what is stopping me from using the API.

shedaniel:
Sorry for the delay! I finally understand how the event is before the client setup.

I am thinking perhaps we can pass in the Supplier of the registry entry instead? And tell people to not use the CLIENT_SETUP event to register particles? I know this will be quite annoying, but I think the API this way would be much more clean then having multiple events for specific hooks.

MaxNeedsSnacks:
Sorry for sort of... necroing this dormant pull request, but after just encountering this myself, I think I just came up with another solution for this with which we could keep the current API structure (apart from having to fix the event bus on Forge obviously) — Inside their client init (**not** the setup), people could use `RegistrySupplier.listen` to register their particles to the engine whenever the entry is registered. Obviously, it should also be easy enough to _add_ a helper to ParticleProviderRegistry that accepts a `RegistrySupplier` and does basically the same thing, this would just mean we wouldn't have to upend the current API structure there

MaxNeedsSnacks:
Superseded by #305, not sure why GitHub didn't automatically close based on the keywords but oh well :/

PR #201, opened 2022-02-15
Some cleanup for Platform class and other platform-specific stuff

MaxNeedsSnacks:


common/src/main/java/dev/architectury/platform/Platform.java
```diff
@@ -60,54 +62,119 @@ public static String getMinecraftVersion() {
...
     
+    /**
+     * Gets a {@link Mod} container by its mod ID.
+     *
+     * @param id The mod ID to look for.
+     * @return The mod container, if found.
+     * @throws NoSuchElementException if no mod with the given ID exists
+     */
     @ExpectPlatform
+    @NotNull
```
shedaniel: We should just not have @NotNulls

shedaniel (approved):

PR #202, opened 2022-02-16
Add way for other mods to indicate that they are fake players

### LINKED ISSUE #30: Fake Players
(from the [The Big, Fat Update ™️](https://github.com/MaxNeedsSnacks/FTB-Ultimine/commit/00b9188e266f890364dd8733bdd978311a798673) commit)

If there's such a thing as a fake player in any commonly used Fabric mods (or FAPI itself), we should probably wrap that using Architectury. If not, we could make our own and maybe have the Forge fake player class be compatible with it somehow?

### LINKED ISSUE #261: PlayerHooks.isFake always returns false on fabric
FTB Teams uses this hook to check if a player is fake. Create uses [Fake Player API](https://github.com/CafeteriaGuild/fake-player-api) for the fake players of several blocks. Since the result is always false, FTB Teams believes the players are not fake, and crashes.
Create issue: https://github.com/Fabricators-of-Create/Create/issues/307

shedaniel:


fabric/src/main/java/dev/architectury/hooks/level/entity/fabric/PlayerHooksImpl.java
```diff
@@ -23,6 +23,6 @@
 
 public class PlayerHooksImpl {
     public static boolean isFake(Player player) {
-        return false;
+        return !FakePlayers.EVENT.invoker().isFakePlayer(player).isTrue();
```
MaxNeedsSnacks: 1. The logic seems a bit backwards on this: A player should be considered fake if this event handler returns `false`?
2. As outlined in [this comment](https://github.com/architectury/architectury-api/issues/261#issuecomment-1133887562), I feel like a reasonable default if no boolean result is given (i.e. `pass`) should be to check whether the player extends `ServerPlayer`, but themselves *isn't* a ServerPlayer

MaxNeedsSnacks (approved):

Technici4n (approved):

MaxNeedsSnacks:
You know, I *should've* bumped versions there, but to be honest, this is a small enough change that it doesn't *quite* warrant that haha

PR #203, opened 2022-02-16
Mark OptiFabric as incompatible

shedaniel:
This will be merged February 26, 2022 if OptiFabric have not made an update to fix this.


MaxNeedsSnacks:
Adding last call to indicate this is time-sensitive

PR #235, opened 2022-04-07
Synchronise it properly?

### LINKED ISSUE #135: Server crashes sometimes with "Where is architectury?"
Hi,

I'm working on a custom modpack and thiss mod is currently a sore thumb of mine...
Most of the time, everything works perfectly, but every so often (once in 20 server starts or something like that), It just crashes, throwing an "IllegalStateException: Where is architectury?"

[crash report](https://pastebin.com/yeE0HCpg)

I've tried to have a look at the code to figure out, what goes wrong, but I was unable to find anything.

I'd love a fix or simply an explaination of how this can happen, so I can try to debug my setup further.

shedaniel:


MaxNeedsSnacks:
Any reason why we couldn't just synchronise the maps themselves, or even the methods? The performance penalty seems negligible, given that this is basically exclusively going to be done at startup and the chance of two mods actually colliding (and thus having to wait for the semaphore here) *seems* to be low enough as well

MaxNeedsSnacks:
TODO: Backport to 1.16

PR #237, opened 2022-04-10
Update MenuRegistryImpl to Fabric API 0.50.0

Juuxel:


shedaniel (approved):
It's fine we can just merge

PR #242, opened 2022-04-11
Close #236

### LINKED ISSUE #236: [Request] Flammable/Fire block registry
Hello, there is currently not a common way provided by the API to make a custom block flammable (so it can not burn or catch fire). With fabric for 1.18.2 i found out, that there is the `FlammableBlockRegistry`, for forge i am not sure. Is it possible to add support for it please?

shedaniel:
Implements the flammable registry to allow mods register their blocks as flammable, **this PR is currently not tested, test mod pending!**

I am not sure if the fabric flammable registry's default instance is applicable to all fire blocks, probably not, so that needs to be look into and tested as well.

I am wary of the performance issues by looking up the near by fire block, ~~but vanilla fires are already laggy~~, and I can't think of a better way to implement this.

We need to inject to an interface, thus the custom ASM we have here.

### DIFF
--- a/common/src/main/java/dev/architectury/registry/level/block/BlockFlammabilityRegistry.java
+++ b/common/src/main/java/dev/architectury/registry/level/block/BlockFlammabilityRegistry.java
@@ -57,30 +57,6 @@ public final class BlockFlammabilityRegistry {
     }
     
     /**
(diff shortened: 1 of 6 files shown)


Faboslav:
Is there anything that i can do to help with this?

PR #243, opened 2022-04-15
Introduce FluidBucketHooks.getFluid

shedaniel:
This requires special handling since forge defers the fiuid, and you can't AT / AW the content field of the bucket item anyways since forge uses asm to redirect calls from the field.

### DIFF
--- a/common/src/main/java/dev/architectury/hooks/fluid/FluidBucketHooks.java
+++ b/common/src/main/java/dev/architectury/hooks/fluid/FluidBucketHooks.java
@@ -31,7 +31,6 @@ public final class FluidBucketHooks {
      * @param item the bucket item
      * @return the fluid contained in the bucket
      */
-    
     @ExpectPlatform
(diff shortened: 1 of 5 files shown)


MaxNeedsSnacks (approved):

common/src/main/java/dev/architectury/hooks/fluid/FluidBucketHooks.java
```diff
@@ -0,0 +1,39 @@
...
+public final class FluidBucketHooks {
+    /**
+     * Returns the fluid contained in the bucket.
+     * This requires special handling since forge defers the fiuid.
+     *
+     * @param item the bucket item
+     * @return the fluid contained in the bucket
+     */
+    
+    @ExpectPlatform
```
Juuxel:
```suggestion
     */
    @ExpectPlatform
```
weird empty line

Juuxel (approved):

PR #257, opened 2022-05-08
Add simple interface injection extensions

shedaniel:
This PR adds `arch$holder()` to many frequent classes: `Item`, `Block`, `GameEvent`, `Fluid`, and `EntityType`.

This also adds `arch$effect` for Food.Properties builder and `arch$getFluid` into `BucketItem`.

The scope of the PR can be extended in the future, but I think this is enough for now.

We are not an API used by literally every mod yet, which is why the methods are prefixed with `arch$` to avoid conflicts.

### DIFF
--- a/common/src/main/java/dev/architectury/extensions/injected/InjectedBlockEntityTypeExtension.java
+++ b/common/src/main/java/dev/architectury/extensions/injected/InjectedBlockEntityTypeExtension.java
@@ -1,0 +1,32 @@
+/*
+ * This file is part of architectury.
+ * Copyright (C) 2020, 2021, 2022 architectury
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+package dev.architectury.extensions.injected;
+
+import dev.architectury.registry.registries.Registries;
+import net.minecraft.core.Registry;
+import net.minecraft.resources.ResourceLocation;
+import net.minecraft.world.level.block.entity.BlockEntityType;
+
+public interface InjectedBlockEntityTypeExtension extends InjectedRegistryEntryExtension {
+    @Override
+    default ResourceLocation arch$getRegistryName() {
+        return Registries.getId((BlockEntityType<?>) this, Registry.BLOCK_ENTITY_TYPE_REGISTRY);
+    }
+}
--- a/common/src/main/java/dev/architectury/extensions/injected/InjectedBlockExtension.java
+++ b/common/src/main/java/dev/architectury/extensions/injected/InjectedBlockExtension.java
@@ -20,13 +20,13 @@
 package dev.architectury.extensions.injected;
 
 import dev.architectury.registry.registries.Registries;
+import net.minecraft.core.Holder;
 import net.minecraft.core.Registry;
-import net.minecraft.resources.ResourceLocation;
 import net.minecraft.world.level.block.Block;
 
-public interface InjectedBlockExtension extends InjectedRegistryEntryExtension {
+public interface InjectedBlockExtension extends InjectedRegistryEntryExtension<Block> {
     @Override
-    default ResourceLocation arch$getRegistryName() {
-        return Registries.getId((Block) this, Registry.BLOCK_REGISTRY);
+    default Holder<Block> arch$holder() {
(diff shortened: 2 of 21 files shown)


common/src/main/java/dev/architectury/extensions/injected/InjectedBlockEntityTypeExtension.java
```diff
@@ -0,0 +1,32 @@
...
+import dev.architectury.registry.registries.Registries;
+import net.minecraft.core.Registry;
+import net.minecraft.resources.ResourceLocation;
+import net.minecraft.world.level.block.entity.BlockEntityType;
+
+public interface InjectedBlockEntityTypeExtension extends InjectedRegistryEntryExtension {
+    @Override
+    default ResourceLocation arch$getRegistryName() {
+        return Registries.getId((BlockEntityType<?>) this, Registry.BLOCK_ENTITY_TYPE_REGISTRY);
+    }
```
MaxNeedsSnacks: In my opinion this should be using the builtInRegistryHolder, given that that is why we wanted to add this extension in the first place

(My originally proposed structure was going to be a `WithBuiltInHolder<T>` interface which just relies on the already existing builtInRegistryHolder method, and that would then be used for methods such as getRegistryKey, which would return a `ResourceKey<T>`, and getRegistryName, which would return the ResourceLocation)

common/src/main/java/dev/architectury/extensions/injected/InjectedBlockExtension.java
```diff
@@ -0,0 +1,32 @@
...
+
+import dev.architectury.registry.registries.Registries;
+import net.minecraft.core.Registry;
+import net.minecraft.resources.ResourceLocation;
+import net.minecraft.world.level.block.Block;
+
+public interface InjectedBlockExtension extends InjectedRegistryEntryExtension {
+    @Override
+    default ResourceLocation arch$getRegistryName() {
+        return Registries.getId((Block) this, Registry.BLOCK_REGISTRY);
```
MaxNeedsSnacks: (Same as for InjectedBlockEntityTypeExtension)

common/src/main/java/dev/architectury/extensions/injected/InjectedEntityTypeExtension.java
```diff
@@ -0,0 +1,32 @@
...
+
+import dev.architectury.registry.registries.Registries;
+import net.minecraft.core.Registry;
+import net.minecraft.resources.ResourceLocation;
+import net.minecraft.world.entity.EntityType;
+
+public interface InjectedEntityTypeExtension extends InjectedRegistryEntryExtension {
+    @Override
+    default ResourceLocation arch$getRegistryName() {
+        return Registries.getId((EntityType<?>) this, Registry.ENTITY_TYPE_REGISTRY);
```
MaxNeedsSnacks: (Same as for InjectedBlockEntityTypeExtension)

common/src/main/java/dev/architectury/extensions/injected/InjectedFluidExtension.java
```diff
@@ -0,0 +1,32 @@
...
+
+import dev.architectury.registry.registries.Registries;
+import net.minecraft.core.Registry;
+import net.minecraft.resources.ResourceLocation;
+import net.minecraft.world.level.material.Fluid;
+
+public interface InjectedFluidExtension extends InjectedRegistryEntryExtension {
+    @Override
+    default ResourceLocation arch$getRegistryName() {
+        return Registries.getId((Fluid) this, Registry.FLUID_REGISTRY);
```
MaxNeedsSnacks: (Same as for InjectedBlockEntityTypeExtension)

common/src/main/java/dev/architectury/extensions/injected/InjectedItemExtension.java
```diff
@@ -0,0 +1,32 @@
...
+
+import dev.architectury.registry.registries.Registries;
+import net.minecraft.core.Registry;
+import net.minecraft.resources.ResourceLocation;
+import net.minecraft.world.item.Item;
+
+public interface InjectedItemExtension extends InjectedRegistryEntryExtension {
+    @Override
+    default ResourceLocation arch$getRegistryName() {
+        return Registries.getId((Item) this, Registry.ITEM_REGISTRY);
```
MaxNeedsSnacks: (Same as for InjectedBlockEntityTypeExtension)

MaxNeedsSnacks:
Should be stated with the above comments that generics in injected interfaces currently do not work properly from what I have heard (https://github.com/FabricMC/fabric-loom/issues/597), and modmuss is aiming to rewrite mixin compile extensions to use ASM rather than reflection, so we might wanna wait for that

MaxNeedsSnacks:
RE the mixin remapping issues, a PR is in the works to hopefully fix these \o/ https://github.com/FabricMC/fabric-mixin-compile-extensions/pull/19

Edit: Whoops, it's already been merged, that's even better ^^

MaxNeedsSnacks (approved):
We might be able to get rid of the "arch$" prefix if we find there to be no conflicting implementations (slash if iface injection actually deals with that kind of stuff smoothly), but for now is looks good to me!

shedaniel:
The issue is that it would be hard to change if there is _actually_ a conflict we find a month later, as seen with quilt QSL, they were able to back down since they are in beta phase, we can't do that.

MaxNeedsSnacks:
"20 days ago" hahahaha *oops*

(please merge this in with the next feature pull request)

PR #268, opened 2022-05-27
Further platform extensions

MaxNeedsSnacks:
This adds two methods to Platform, `Platform.getLoader()` which is a simple wrapper around `ArchitecturyTarget` for mods not using the plugin, and `Platform.getFilePaths()`, which can be used to get *all* root file paths of a mod that is split into multiple source sets, for instance.

Signed-off-by: Max <maxh2709@gmail.com>

### DIFF
--- a/common/src/main/java/dev/architectury/platform/Platform.java
+++ b/common/src/main/java/dev/architectury/platform/Platform.java
@@ -51,10 +51,14 @@ public final class Platform {
             return;
         }
         
-        switch (ArchitecturyTarget.getCurrentTarget()) {
+        switch (getLoader()) {
             case "fabric" -> simpleLoaderCache = 0;
             case "forge" -> simpleLoaderCache = 1;
         }
+    }
+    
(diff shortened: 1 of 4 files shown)


common/src/main/java/dev/architectury/platform/Platform.java
```diff
@@ -51,12 +51,16 @@ private static void updateLoaderCache() {
...
         }
         
-        switch (ArchitecturyTarget.getCurrentTarget()) {
+        switch (getLoader()) {
             case "fabric" -> simpleLoaderCache = 0;
             case "forge" -> simpleLoaderCache = 1;
         }
     }
     
+    public static String getLoader() {
```
shedaniel: We shouldn't expose this, ArchitecturyTarget.getCurrentTarget() is the compiled target, so on quilt it is still fabric because this is compiled for fabric. This has quite specific implications.

shedaniel (approved):

PR #272, opened 2022-06-09
Implement hasTag onto BiomeContext

### LINKED ISSUE #271: BiomeSelectionContext wrapper does not expose necessary information
The BiomeModification API currently does not expose many methods from the underlying Fabric `BiomeSelectionContext` that are necessary for the API's use. For example, I wish to filter a feature by a biome tag, but this is not possible because it requires registry access and the modifications API calls its checks before the server is fully constructed and thus accessible for registry access.

The API should expose `boolean hasTag(TagKey<Biome> tag);` or `getHolder` via `BiomeContext`, or even the `Biome` instance itself. As of current `BiomeContext` does not include enough data for a lot of things.

itsmeow:
Untested, probably works. Really hacky solution for forge, could be improved?

Resolves #271

### DIFF
--- a/forge/src/main/java/dev/architectury/registry/level/biome/forge/BiomeModificationsImpl.java
+++ b/forge/src/main/java/dev/architectury/registry/level/biome/forge/BiomeModificationsImpl.java
@@ -30,6 +30,7 @@ import net.minecraft.core.Holder;
 import net.minecraft.core.Registry;
 import net.minecraft.resources.ResourceKey;
 import net.minecraft.resources.ResourceLocation;
+import net.minecraft.server.MinecraftServer;
 import net.minecraft.sounds.Music;
 import net.minecraft.sounds.SoundEvent;
 import net.minecraft.tags.TagKey;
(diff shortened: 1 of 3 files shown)


shedaniel:
was this tested?

itsmeow:
> Untested

shedaniel:
yeah do that 😅

itsmeow:
Tested it, seems to work on both implementations.

forge/src/main/java/dev/architectury/registry/level/biome/forge/BiomeModificationsImpl.java
```diff
@@ -136,6 +139,11 @@ public Optional<ResourceLocation> getKey() {
             public BiomeProperties getProperties() {
                 return properties;
             }
+            
+            @Override
+            public boolean hasTag(TagKey<Biome> tag) {
+                return GameInstance.getServer().registryAccess().registryOrThrow(Registry.BIOME_REGISTRY).getHolder(biomeResourceKey.get()).get().is(tag);
```
shedaniel: Would it be wise here to check if the `getHolder` optional is null?

itsmeow: Maybe, but it'll error out anyway, so I don't see the point. The chances of it being null are near zero.

itsmeow:
Should be good to merge now.

shedaniel (approved):
Some styling issues, but we will fix that on merge, LGTM

PR #276, opened 2022-06-13
Add CommandBuildContext to CommandRegistrationEvent (breaking, thanks to Forge <3)

MaxNeedsSnacks:
What a wonderful tiny change that'll break **every single downstream mod**

We should really look into overhauling events for 1.20, that discussion is gonna follow soon!


MaxNeedsSnacks:
Likely deferring this to 1.19.1 *at least* as introducing breaking changes like these during a release cycle is generally considered bad practice ~~unless you already have a chokehold on the market anyways~~ and there is an easy enough workaround for people that should work for most use cases (using the builtin registry access)

MaxNeedsSnacks:
Superseded by #302

PR #277, opened 2022-06-14
Add stack-aware CraftingRemainingItem

shedaniel:


Technici4n (approved):

common/src/main/java/dev/architectury/hooks/item/ItemStackHooks.java
```diff
@@ -53,4 +54,14 @@ public static void giveItem(ServerPlayer player, ItemStack stack) {
...
+    
+    @ExpectPlatform
+    public static boolean hasCraftingRemainingItem(ItemStack stack) {
+        throw new AssertionError();
+    }
+    
+    @ExpectPlatform
+    public static ItemStack getCraftingRemainingItem(ItemStack stack) {
+        throw new AssertionError();
+    }
```
MaxNeedsSnacks: Should probably be mentioned that these are current **not** stack-aware on Fabric...

MaxNeedsSnacks (approved):

PR #280, opened 2022-06-15
Implement Forge's new Fluid API

MaxNeedsSnacks:
The new Forge Fluid API, introduced in build `41.0.28`, is a breaking change that directly affects Architectury as we used to rely on their FluidAttributes class for our own fluid attributes system. This PR updates, in as direct and non-breaking of a way as possible, that system to support Forge's new `FluidType` system instead.

TODOs:
- [ ] Test in userdev; this is unfortunately not possible for *me* to do because of a weird cyclic dependency error I seem to be the only one getting
- [ ] Add state-aware getters to FluidStackHooks and deprecate the getters using `Level` and `BlockPos` but not FluidState for removal
- [ ] The same should be applied to ArchitecturyFluidAttributes as well; maybe we should look into completely separating stack-aware and state-aware getters.
- [ ] General code cleanup
- [ ] At least look into adding support for certain new features in Forge's Fluid API if a Fabric-equivalent exists or is easy to create ourselves

### DIFF
--- a/common/src/main/java/dev/architectury/core/fluid/ArchitecturyFluidAttributes.java
+++ b/common/src/main/java/dev/architectury/core/fluid/ArchitecturyFluidAttributes.java
@@ -343,6 +343,7 @@ public interface ArchitecturyFluidAttributes {
      * @deprecated Please use and override {@link #getColor(FluidState, BlockAndTintGetter, BlockPos)}
      * or {@link #getColor(FluidStack)} instead, this method will be removed in a future version.
      */
+    @Deprecated(forRemoval = true)
     int getColor(@Nullable FluidStack stack, @Nullable BlockAndTintGetter level, @Nullable BlockPos pos);
     
     /**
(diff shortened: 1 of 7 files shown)


MaxNeedsSnacks:
Looks like I forgot about ArchitecturyFlowingFluid, give me a second

shedaniel (approved):
I will leave this in final call for another half day, then we will merge it because this is high priority

common/src/main/java/dev/architectury/core/fluid/ArchitecturyFluidAttributes.java
```diff
@@ -258,16 +293,70 @@ default ResourceLocation getFlowingTexture() {
...
      * Returns the color of the fluid.
      *
      * @param stack the fluid stack, can be {@code null}
      * @param level the level, can be {@code null}
      * @param pos   the position, can be {@code null}
      * @return the color
+     * @deprecated Please use and override {@link #getColor(FluidState, BlockAndTintGetter, BlockPos)}
+     * or {@link #getColor(FluidStack)} instead, this method will be removed in a future version.
      */
     int getColor(@Nullable FluidStack stack, @Nullable BlockAndTintGetter level, @Nullable BlockPos pos);
```
Juuxel: This is missing the annotation:
```suggestion
     * @deprecated Please use and override {@link #getColor(FluidState, BlockAndTintGetter, BlockPos)}
     * or {@link #getColor(FluidStack)} instead, this method will be removed in a future version.
     */
    @Deprecated(forRemoval = true)
    int getColor(@Nullable FluidStack stack, @Nullable BlockAndTintGetter level, @Nullable BlockPos pos);
```

PR #287, opened 2022-06-28
Add loot table modification event

### LINKED ISSUE #42: Loot Table injecting
We need to be able to inject into vanilla (or other modded) loot tables using some sort of event or by offering some special JSON file syntax.

This would be used by most mods that add loot to stuff like dungeon chests, or that want to add to things like block drops (e.g. extra seeds from grass)

Juuxel:
Closes #42. It's a simple wrapper around the platform events.

(this is for 1.19 but should be the exact same when backporting)

### DIFF
--- a/common/src/main/java/dev/architectury/event/events/common/LootEvent.java
+++ b/common/src/main/java/dev/architectury/event/events/common/LootEvent.java
@@ -1,0 +1,87 @@
+/*
+ * This file is part of architectury.
+ * Copyright (C) 2020, 2021, 2022 architectury
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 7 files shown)


common/src/main/java/dev/architectury/event/events/common/LootEvent.java
```diff
@@ -0,0 +1,87 @@
...
+    @FunctionalInterface
+    interface ModifyLootTable {
+        /**
+         * Modifies a loot table.
+         *
+         * @param lootTables the {@link LootTables} instance containing all loot tables
+         * @param id         the loot table ID
+         * @param context    the context used to modify the loot table
+         */
+        void modifyLootTable(LootTables lootTables, ResourceLocation id, LootTableModificationContext context);
```
shedaniel: Would be nice as per discussed previously the `isBuiltin` param

Juuxel: Yeah I've just forgotten about this pr, I'll fix it up today

Juuxel: Okay done, also rebased onto master

shedaniel (approved):

MaxNeedsSnacks (approved):

MaxNeedsSnacks:
Something to consider for the future: Implement non-builtin loot table modification on Forge, either through a mixin or by requesting (or PRing) this as a feature on their end

PR #299, opened 2022-07-16
Fix chat events

shedaniel:
- This fixes chat events on fabric due to Fabric API taking priority above our mixins, we inject to the same place
- Test mod now with appended "new text" for server chat modifications and "error:" for client chat modifications
- Forge event now not cancelled for client chat modifications


fabric/src/main/java/dev/architectury/impl/fabric/EventChatDecorator.java
```diff
@@ -19,68 +19,8 @@
...
-import net.minecraft.network.chat.PlayerChatMessage;
-import net.minecraft.server.level.ServerPlayer;
-import net.minecraft.server.network.FilteredText;
-import org.jetbrains.annotations.Nullable;
 
-import java.util.concurrent.CompletableFuture;
-
-public class EventChatDecorator implements ChatDecorator {
+public class EventChatDecorator {
     public static final Component CANCELLING_COMPONENT = Component.literal("THIS SHOULDN'T BE DISPLAYED, ARCHITECTURY SPECIFIC STRING DO NOT IMITATE THIS");
```
MaxNeedsSnacks, line 25: Is there really no other way to "cancel" the player's chat message than to send a weird special component? For instance, what happens if the Future returns null or completes exceptionally?

MaxNeedsSnacks:
Superseded by #302

shedaniel:
We still need this

MaxNeedsSnacks:
In that case, see my comments regarding the cancelling component ^^ There *has* to be a better way to do that :/

fabric/src/main/java/dev/architectury/impl/fabric/EventChatDecorator.java
shedaniel: Not really, this is a compromise for this badly designed event

MaxNeedsSnacks: Honestly, yeah, then it's... fine, considering 1.19 has basically been superseded by 1.19.2+ anyways

MaxNeedsSnacks (approved):

MaxNeedsSnacks:
Please resolve the merge conflicts so this can be put on last call

MaxNeedsSnacks:
Is there still enough demand for 1.19 to warrant fixing this issue? The discourse(TM) regarding that version has largely died down, and due to CurseForge's horrible UX I would even say that publishing a new *1.19.0* release would likely confuse users into downloading said new version for their *1.19.2* instance

PR #305, opened 2022-07-30
Fix ParticleProviderRegistry not working properly on Forge, add Javadocs

MaxNeedsSnacks:
The long and short of it is that we (a) were not listening to Forge's particle factory event on the correct bus and that (b) users were using the `CLIENT_SETUP` event to register their factories, which as mentioned in #196 is after the particle factory event has been fired on Forge.

Based on https://github.com/architectury/architectury-api/pull/196#issuecomment-1200268981, this supersedes and thus closes #196.


shedaniel (approved):
LGTM, might be a good call to have a logger warn to the logs that it may be too late to register? If that's possible

MaxNeedsSnacks:
Well, seeing as everything should be done before or during that event on Forge, we could just check if the list is null and print a warning in that case, right?

shedaniel (approved):

Issue #324, opened 2022-08-21
Forge 43.1.1 (1.19.2) - Crash when using chat


big-julius:
Hello,

When anyone on my server try to post a message in the chat : all players connected have their game crashing.
I send the crash report.
[crash-2022-08-21_12.00.30-client.txt](https://github.com/architectury/architectury-api/files/9388838/crash-2022-08-21_12.00.30-client.txt)

Quipex:
I confirm that, the exception occurs on client

```
java.lang.NoSuchMethodError: 'net.minecraft.network.chat.ChatType net.minecraftforge.client.event.ClientChatReceivedEvent.getType()'
	at dev.architectury.event.forge.EventHandlerImplClient.event(EventHandlerImplClient.java:120) ~[architectury-5.12.42-forge.jar%23147!/:?] {re:classloading,pl:runtimedistcleaner:A}
```

Juuxel:
On Minecraft 1.19.1 and 1.19.2 you have to use Architectury 6.x, but you have 5.x. Upgrading should fix it.

Quipex:
Thanks! I've downloaded 5.x because it was on top
![image](https://user-images.githubusercontent.com/30311108/185797585-4de1637a-8c22-4159-84aa-13818174db90.png)

big-julius:
Oh sh*t! I did the same mistake... Sorry and thankk you for the reply. ;)

PR #325, opened 2022-08-22
Prevent running 1.19 Arch API on 1.19.1+

Juuxel:
Should reduce the number of issues and support threads like #324.


Juuxel:
While not the most descriptive, these screens at least show what the actual issue is (instead of an obscure mixin/chat crash) and will probably get people to look for the correct versions.

Fabric dependency screen
Forge dependency screen

MaxNeedsSnacks (approved):

shedaniel (approved):

PR #341, opened 2022-09-17
Fix #340

### LINKED ISSUE #340: Identifiable ResourceReloadListener and ordering
~~Thank you GitHub for wrongly submitting the issue early
So, hopefully I can write quickly enough before anyone sees this incomplete issue.~~

Architectury API allows to register resource reload listeners via [ReloadListenerRegistry](https://github.com/architectury/architectury-api/blob/6a9d61aa425fec0ea24be8a56b14b9340f23cf55/common/src/main/java/dev/architectury/registry/ReloadListenerRegistry.java#L31). This is neat up until you see the Fabric/Quilt's take on the system.

In the Fabric and Quilt ecosystem those reloaders are named and have stable identifiers, this is important as it allows other reloaders to re-order them (depend or make sure to run before, etc).
This can be very important in the case of addon mods, and others.

This expectation is entirely wrecked with Architectury API not letting mods name their reloaders and giving them random IDs.
In addition, it doesn't let anyone re-order reloaders, which led to some people trying to order using registration call order (and that is entirely wrecked by Quilt's own reloader sorter since it's not using Lists but HashMaps).

It would be nice to add a way to give stable names and specify dependencies on resource reloaders inside of Architectury API to satisfy Fabric and Quilt's expectations.

shedaniel:
ReloadListenerRegistry now with optional RSLoc argument and dependencies argument, we are not going to sort the dependencies on Forge, that's too much of a headache.

### DIFF
--- a/common/src/main/java/dev/architectury/registry/ReloadListenerRegistry.java
+++ b/common/src/main/java/dev/architectury/registry/ReloadListenerRegistry.java
@@ -20,15 +20,28 @@
 package dev.architectury.registry;
 
 import dev.architectury.injectables.annotations.ExpectPlatform;
+import net.minecraft.resources.ResourceLocation;
 import net.minecraft.server.packs.PackType;
 import net.minecraft.server.packs.resources.PreparableReloadListener;
+
+import javax.annotation.Nullable;
+import java.util.Collection;
+import java.util.List;
 
 public final class ReloadListenerRegistry {
     private ReloadListenerRegistry() {
     }
     
-    @ExpectPlatform
     public static void register(PackType type, PreparableReloadListener listener) {
+        register(type, listener, null);
+    }
+    
+    public static void register(PackType type, PreparableReloadListener listener, @Nullable ResourceLocation listenerId) {
+        register(type, listener, listenerId, List.of());
+    }
+    
+    @ExpectPlatform
+    public static void register(PackType type, PreparableReloadListener listener, @Nullable ResourceLocation listenerId, Collection<ResourceLocation> dependencies) {
         throw new AssertionError();
     }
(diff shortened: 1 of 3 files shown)


common/src/main/java/dev/architectury/registry/ReloadListenerRegistry.java
```diff
@@ -20,15 +20,28 @@
 package dev.architectury.registry;
 
 import dev.architectury.injectables.annotations.ExpectPlatform;
+import net.minecraft.resources.ResourceLocation;
 import net.minecraft.server.packs.PackType;
 import net.minecraft.server.packs.resources.PreparableReloadListener;
 
+import javax.annotation.Nullable;
```
Juuxel, line 27: Do we use javax nullable?

MaxNeedsSnacks: We do not, please switch to jbr annotations if possible

MaxNeedsSnacks: Gonna fix this in an extra PR after this one actually, since we do seem to be using javax in some other places where we shouldn't

MaxNeedsSnacks (approved):
Side note: IMO users should always supply an ID for their reload listeners for proper ordering on Fabriquilt, so maybe we should deprecate the method without the `listenerId` parameter for removal in a future version of the API (as well as not mark the parameter as explicitly nullable if that deprecation is introduced)

I'm happy to merge this in its current state, however, as it is more important that we *get this fixed* than to have the API laid out perfectly, and this has been dormant for too long anyways :P

PR #348, opened 2022-10-30
Fix Loom 1.0 removals

Juuxel:
- javax annotations were accidentally available because of a transitive dep of Minecraft. Loom 0.13 blocks them since the launcher doesn't download them either.
- `forge.localMods` DSL was removed in favour of just `mods`


shedaniel (approved):

MaxNeedsSnacks:
Something tells me we forgot to merge this, and now Loom 1.0 is out so I think it's high time to do so whoops

PR #352, opened 2022-11-19
Fix injection point on lightning strike event

MaxNeedsSnacks:
So, as it turns out, the point at which we currently inject for our lightning strike event means that the entities have already been marked as hit by it (and potentially gotten hurt / died), and the list of entities instead consists of all entities within a 15 block radius *not* yet hit by lightning(?), which both seems counterproductive and does not match the description of the corresponding event.

By moving the injection point to the *second* getEntities call, we get the desired result, though as a side effect the event will be called multiple times in a lightning bolt's lifetime (well, twice, as they live for two ticks before dying)


MaxNeedsSnacks:
Note: A clean merge to 1.19 is possible, and I'll see about doing that once merged ^^

shedaniel (approved):

PR #370, opened 2022-12-23
Add creative tab access wideners

### LINKED ISSUE #369: [1.19.3] Add CreativeModeTabs fields to default access transformer
The `CreativeModeTabs` class has vanilla's creative tabs as private static variables, in order to add an item to an existing tab you would have to access transform it, which it seems like something Architectury should already do.

itsmeow:
Resolves #369 

Allows easy use of vanilla tabs with `arch$tab()`

image

### DIFF
--- a/common/src/main/resources/architectury.accessWidener
+++ b/common/src/main/resources/architectury.accessWidener
@@ -139,6 +139,20 @@ accessible class net/minecraft/client/particle/ParticleEngine$MutableSpriteSet
 accessible field net/minecraft/client/particle/ParticleEngine$MutableSpriteSet sprites Ljava/util/List;
 transitive-accessible class net/minecraft/world/item/CreativeModeTab$Output
 transitive-accessible class net/minecraft/world/item/CreativeModeTab$TabVisibility
+transitive-accessible field net/minecraft/world/item/CreativeModeTabs BUILDING_BLOCKS Lnet/minecraft/world/item/CreativeModeTab;
+transitive-accessible field net/minecraft/world/item/CreativeModeTabs FUNCTIONAL_BLOCKS Lnet/minecraft/world/item/CreativeModeTab;
+transitive-accessible field net/minecraft/world/item/CreativeModeTabs REDSTONE_BLOCKS Lnet/minecraft/world/item/CreativeModeTab;
+transitive-accessible field net/minecraft/world/item/CreativeModeTabs HOTBAR Lnet/minecraft/world/item/CreativeModeTab;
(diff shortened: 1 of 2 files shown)


common/src/main/resources/architectury.accessWidener
```diff
@@ -139,6 +139,20 @@ accessible class net/minecraft/client/particle/ParticleEngine$MutableSpriteSet
...
+transitive-accessible field net/minecraft/world/item/CreativeModeTabs COMBAT Lnet/minecraft/world/item/CreativeModeTab;
+transitive-accessible field net/minecraft/world/item/CreativeModeTabs SPAWN_EGGS Lnet/minecraft/world/item/CreativeModeTab;
+transitive-accessible field net/minecraft/world/item/CreativeModeTabs INVENTORY Lnet/minecraft/world/item/CreativeModeTab;
+transitive-accessible field net/minecraft/world/item/CreativeModeTabs NATURAL_BLOCKS Lnet/minecraft/world/item/CreativeModeTab;
+transitive-accessible field net/minecraft/world/item/CreativeModeTabs COLORED_BLOCKS Lnet/minecraft/world/item/CreativeModeTab;
+transitive-accessible field net/minecraft/world/item/CreativeModeTabs TOOLS_AND_UTILITIES Lnet/minecraft/world/item/CreativeModeTab;
+transitive-accessible field net/minecraft/world/item/CreativeModeTabs FOOD_AND_DRINKS Lnet/minecraft/world/item/CreativeModeTab;
+transitive-accessible field net/minecraft/world/item/CreativeModeTabs INGREDIENTS Lnet/minecraft/world/item/CreativeModeTab;
+transitive-accessible field net/minecraft/world/item/CreativeModeTabs OP_BLOCKS Lnet/minecraft/world/item/CreativeModeTab;
```
triphora: Wouldn't it be better to make these autogenerated?

itsmeow: Not sure how to do that

itsmeow: I see the generateAccessWideners but I'm not good with Groovy at all or bytecode stuff

MaxNeedsSnacks: I'll write something up tomorrow. We'd basically just have to iterate over all the fields of the `net/minecraft/world/item/CreativeModeTabs` class node (which we can easily do, see the generateRenderTypeRelated method) and then check if its type is of `Lnet/minecraft/world/item/CreativeModeTab;` (which we can also easily do)

itsmeow: Added generation

shedaniel (approved):
LGTM

MaxNeedsSnacks (approved):

triphora (approved):

PR #379, opened 2023-01-25
Add support for adding PacketTransformers to SimpleNetworkManager messages

MaxNeedsSnacks:
Using this, developers will be able to use SNM to create messages that, for example, are split into smaller packets (which may be useful for large bits of information like a chat message history, which would otherwise break the packet length limit)

_Can be ported without changes to 1.19.x_

### DIFF
--- a/common/src/main/java/dev/architectury/networking/simple/SimpleNetworkManager.java
+++ b/common/src/main/java/dev/architectury/networking/simple/SimpleNetworkManager.java
@@ -20,9 +20,12 @@
 package dev.architectury.networking.simple;
 
 import dev.architectury.networking.NetworkManager;
+import dev.architectury.networking.transformers.PacketTransformer;
 import dev.architectury.platform.Platform;
 import dev.architectury.utils.Env;
 import net.minecraft.resources.ResourceLocation;
+
+import java.util.List;
 
 /**
  * A simple wrapper for {@link NetworkManager} to make it easier to register messages and send them to clients/servers.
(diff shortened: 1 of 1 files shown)


common/src/main/java/dev/architectury/networking/simple/SimpleNetworkManager.java
```diff
@@ -48,35 +51,55 @@ private SimpleNetworkManager(String namespace) {
...
      *
-     * @param id      a unique ID for the message, must be a valid value for {@link ResourceLocation#getPath}
-     * @param decoder the message decoder for the message
+     * @param id           a unique ID for the message, must be a valid value for {@link ResourceLocation#getPath}
+     * @param decoder      the message decoder for the message
+     * @param transformers a list of packet transformers to apply to the message packet
      * @return a {@link MessageType} describing the registered message
      */
-    public MessageType registerS2C(String id, MessageDecoder<BaseS2CMessage> decoder) {
+    public MessageType registerS2C(String id, MessageDecoder<BaseS2CMessage> decoder, List<PacketTransformer> transformers) {
```
shedaniel, line 76: Please mark this as Experimental, I think that's also the case on the NetworkManager methods, we can reevaluate this in the future

common/src/main/java/dev/architectury/networking/simple/SimpleNetworkManager.java
```diff
@@ -48,35 +51,55 @@ private SimpleNetworkManager(String namespace) {
...
      *
-     * @param id      a unique ID for the message, must be a valid value for {@link ResourceLocation#getPath}
-     * @param decoder the message decoder for the message
+     * @param id           a unique ID for the message, must be a valid value for {@link ResourceLocation#getPath}
+     * @param decoder      the message decoder for the message
+     * @param transformers a list of packet transformers to apply to the message packet
      * @return a {@link MessageType} describing the registered message
      */
-    public MessageType registerC2S(String id, MessageDecoder<BaseC2SMessage> decoder) {
+    public MessageType registerC2S(String id, MessageDecoder<BaseC2SMessage> decoder, List<PacketTransformer> transformers) {
```
shedaniel, line 108: Same Experiential here

shedaniel (changes requested):

common/src/main/java/dev/architectury/networking/simple/SimpleNetworkManager.java
```diff
@@ -49,34 +53,62 @@ private SimpleNetworkManager(String namespace) {
     }
     
     /**
-     * Registers a server -&gt; client message.
+     * Convenience method to register a server -&gt; client message with no packet transformers.
```
shedaniel:
```suggestion
     * Registers a server -&gt; client message with no packet transformers.
```
I don't think we need the description of this being a helper method

common/src/main/java/dev/architectury/networking/simple/SimpleNetworkManager.java
```diff
@@ -49,34 +53,62 @@ private SimpleNetworkManager(String namespace) {
...
-            NetworkManager.registerReceiver(NetworkManager.s2c(), messageType.getId(), receiver);
+            NetworkManager.registerReceiver(NetworkManager.s2c(), messageType.getId(), transformers, receiver);
         }
         
         return messageType;
     }
     
     /**
-     * Registers a client -&gt; server message.
+     * Convenience method to register a client -&gt; server message with no packet transformers.
```
shedaniel:
```suggestion
     * Registers a client -&gt; server message with no packet transformers.
```

shedaniel (approved):

shedaniel:
Please bump minor when release :p

MaxNeedsSnacks:
Gonna hold off on merging this until tomorrow at least so we can lump it together with #380

PR #380, opened 2023-01-26
Add PlayerEvent.ATTACK_ENTITY

desht:
* Fired from Forge's AttackEntityEvent and Fabric's AttackEntityCallback

* When player is about to attack any entity (not just living) with left-click


fabric/src/main/java/dev/architectury/event/fabric/EventHandlerImpl.java
```diff
@@ -35,10 +35,18 @@
...
 import net.fabricmc.fabric.api.event.player.UseItemCallback;
 import net.fabricmc.fabric.api.loot.v2.LootTableEvents;
 import net.minecraft.commands.Commands;
+import net.minecraft.world.InteractionHand;
+import net.minecraft.world.InteractionResult;
+import net.minecraft.world.entity.Entity;
+import net.minecraft.world.entity.player.Player;
+import net.minecraft.world.level.Level;
+import net.minecraft.world.phys.EntityHitResult;
+import org.jetbrains.annotations.Nullable;
```
shedaniel: are all of these imports needed?

desht: No, I'll fix that

shedaniel (approved):
LGTM, thanks ・∀・

MaxNeedsSnacks (approved):

PR #395, opened 2023-03-28
fix: Fabric BlockItem mixin: fire BlockEvent.PLACE at method head

desht:
Currently, the mixin fires the event after the block has been placed


MaxNeedsSnacks (approved):
note to self / maintainers: apply this change to 1.18 as well

PR #405, opened 2023-05-07
add getEntries method to get a collection of RegistrySuppliers

JT122406:
Good for bulk editing entire sets


shedaniel:
Thank you for your contribution, however I have a few issues with this PR:
- You added `getEntries` to the private inner class, which is not accessible outside, also the inner class is for the each entry
- `DeferredRegister` already implements `Iterable<RegistrySupplier<T>>`, you can use a for-loop to access all entries already

PR #406, opened 2023-05-07
Update Mixin jsons and pack_format

JT122406:


fabric/src/main/resources/architectury.mixins.json
```diff
@@ -2,8 +2,8 @@
   "required": true,
   "package": "dev.architectury.mixin.fabric",
   "plugin": "dev.architectury.plugin.fabric.ArchitecturyMixinPlugin",
-  "compatibilityLevel": "JAVA_16",
-  "minVersion": "0.7.11",
+  "compatibilityLevel": "JAVA_167",
```
topia: At least this change should be wrong
```suggestion
  "compatibilityLevel": "JAVA_17",
```

JT122406: lol my bad

PR #412, opened 2023-06-20
Fix tooltip color event

shedaniel:
I removed this event back in the 1.19.3 port since Forge removed its event. Now that Forge has finally added it back (in May), this PR adds the event back.

### About version support
Since Forge has only fixed this issue on 1.19.4, this PR will not be backported to 1.19.3.
Looking at the grand scheme of things, I propose that we move forward with the following:
- API 4: 1.18.2
- ~~API 5: 1.19 - 1.19.1~~ **(drop support here)**
- API 6: 1.19.2
- ~~API 7: 1.19.3~~ **(drop support here)**
- API 8: 1.19.4
- API 9: 1.20 - 1.20.1

### Test Mod
I have added a simple test mod:
image

### DIFF
--- a/common/src/main/java/dev/architectury/impl/TooltipEventColorContextImpl.java
+++ b/common/src/main/java/dev/architectury/impl/TooltipEventColorContextImpl.java
@@ -24,6 +24,7 @@ import org.jetbrains.annotations.ApiStatus;
 
 @ApiStatus.Internal
 public class TooltipEventColorContextImpl implements ClientTooltipEvent.ColorContext {
+    public static ThreadLocal<TooltipEventColorContextImpl> tooltipColorContext = ThreadLocal.withInitial(TooltipEventColorContextImpl::new);
     private int backgroundColor;
     private int outlineGradientTopColor;
     private int outlineGradientBottomColor;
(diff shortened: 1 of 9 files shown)


MaxNeedsSnacks:
Concur'd on dropping any 1.19 versions other than .2 and .4 (might have to reevaluate which of those two we end up keeping another time), other than that I'll give this a review in a bit

common/src/main/java/dev/architectury/impl/TooltipEventColorContextImpl.java
```diff
@@ -24,6 +24,7 @@
 
 @ApiStatus.Internal
 public class TooltipEventColorContextImpl implements ClientTooltipEvent.ColorContext {
+    public static ThreadLocal<TooltipEventColorContextImpl> tooltipColorContext = ThreadLocal.withInitial(TooltipEventColorContextImpl::new);
```
Juuxel: Can this be final?

MaxNeedsSnacks (approved):
yesyes good and best

Juuxel (approved):

Issue #430, opened 2023-08-23
Different result between forge and fabric when trying to register a block to replace vanila one.


qyl27:
When I tried to register a block with id "minecraft:dandelion" (accroding to forge doc and fabricmc reddit), I found different result between architectury api forge and fabric version.  
In forge, it will replace the vanilla one.  
In fabric, it will shows `Attempted to register ID ResourceKey[minecraft:block / minecraft:dandelion] at different raw IDs (147, 1003)! If you're trying to override an item, use .set(), not .register()!`, and then game crash.  

[My project code repo](https://github.com/qyl27/Dandelion/tree/7c69fff65db8552149c2eb7056ee538ce6780075)

qyl27:
I found a solution after my attempt, in [my fork repo](https://github.com/qyl27/architectury-api/commit/32e18f054544ad4814055a1b99fc6dc3baaeb53e). Should I PR it? Or maybe better solution?

Juuxel:
This is an intentional feature of Fabric API to prevent people from accidentally modifying registered content - replacement has to be explicit.

I also feel that hardcoding the `minecraft` namespace is problematic, a more proper solution would simply check if the content has already been registered.

qyl27:
ah, you are right.  
I have updated it. [commit](https://github.com/qyl27/architectury-api/commit/646a614749343eea4fa94363e406995b0f4edd54)

Juuxel:
The code is still incorrect, now it checks if that *entry* has been registered instead of whether that *id* is already taken.

qyl27:
I have not noticed about `contains(ResourceLocation)` yet, is it correct that I use `contains(id)`?

PR #435, opened 2023-09-21
Registry Replacements (#430)

qyl27:
Replace vanilla items with register() on fabric.  
(Tested on my machine.)

### DIFF
--- a/fabric/src/main/java/dev/architectury/registry/registries/fabric/RegistrarManagerImpl.java
+++ b/fabric/src/main/java/dev/architectury/registry/registries/fabric/RegistrarManagerImpl.java
@@ -208,7 +208,7 @@ public class RegistrarManagerImpl {
         @Override
         public <E extends T> RegistrySupplier<E> register(ResourceLocation id, Supplier<E> supplier) {
             var reference = supplier.get();
-            if (getId(reference) != null) {
+            if (contains(id)) {
                 var rawId = getRawId(get(id));
                 Registry.registerMapping((Registry) delegate, rawId, id.toString(), reference);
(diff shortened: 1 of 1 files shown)


MaxNeedsSnacks:
Hmm, honestly I would prefer if registry replacement was **explicit** on both instead of implicit, so maybe we should instead consider making a registerOrOverride method?

qyl27:
> Hmm, honestly I would prefer if registry replacement was **explicit** on both instead of implicit, so maybe we should instead consider making a registerOrOverride method?

Well, maybe that's right.  
I will change it to a explicit method named `registerOrOverride`.

shedaniel:
After some discussion on Discord, I believe we don't want to support registry replacements seeing how NeoForge no longer supports it after the Registry Rework and fabric never encouraged it. Alternative solutions should be used instead, albeit mixins or etc.

Some checks can be added to the forge register() method to warn potential registry replacements instead.

qyl27:
> After some discussion on Discord, I believe we don't want to support registry replacements seeing how NeoForge no longer supports it after the Registry Rework and fabric never encouraged it. Alternative solutions should be used instead, albeit mixins or etc.
> 
> Some checks can be added to the forge register() method to warn potential registry replacements instead.

Ah, okay.

Jab125:
> After some discussion on Discord, I believe we don't want to support registry replacements seeing how NeoForge no longer supports it after the Registry Rework and fabric never encouraged it. Alternative solutions should be used instead, albeit mixins or etc.
> 
> Some checks can be added to the forge register() method to warn potential registry replacements instead.

PR #463, opened 2023-12-24
Add ClientTooltipComponentRegistry for custom tooltips

Prototik:


fabric/src/main/java/dev/architectury/mixin/fabric/client/MixinClientTooltipComponent.java
MaxNeedsSnacks, line 1: Didn't Fabric have a native API for this stuff as well?

common/src/main/java/dev/architectury/registry/client/gui/ClientTooltipComponentManager.java
MaxNeedsSnacks, line 1: Nitpick: This should probably be called ClientTooltipComponentRegistry to match the naming pattern of other classes like this

fabric/src/main/java/dev/architectury/mixin/fabric/client/MixinClientTooltipComponent.java
Prototik: Yeah, it's `TooltipComponentCallback`, didn't found it initially, lemme a second to rewrite it...

PR #467, opened 2023-12-30
Make ArchitecturyFlowingFluid use the same Forge FluidType when the s…

### LINKED ISSUE #447: No way to make ArchitecturyFlowingFluid.Source and ArchitecturyFlowingFluid.Flowing liquids use the same FluidType (ArchitecturyFluidAttributesForge)
Minecraft: 1.19.2
Forge: 43.2.23
Architectury: git, 1.19.2 branch

The Source and Flowing instances for the same fluid should be using the same "FluidType", but the constructors generate new ones. I came up with the following solution, but it's a little... "dirty", but it works. I didn't create a PR because my editor messes with all of the whitespace, and it's just a pain to undo. If you want me to do one anyways, I can. 

I added caches for the Properties and FluidTypes, using the "attributes" object as the key. I'm currently making an addon for KubeJS to do Fluid Interactions (cobblegen, basaltgen), and this is how I found this, as that system makes heavy use of FluidTypes. I left out the imports and the methods I didn't touch. 

```
public abstract class ArchitecturyFlowingFluid extends ForgeFlowingFluid {
    private final ArchitecturyFluidAttributes attributes;
    private final Supplier<FluidType> forgeType;
    private static final Map<ArchitecturyFluidAttributes, Properties> cachedProperties;
    private static final Map<ArchitecturyFluidAttributes, FluidType> cachedFluidType;
    static {
        cachedProperties = new HashMap<>();
        cachedFluidType = new HashMap<>();
    }

    ArchitecturyFlowingFluid(ArchitecturyFluidAttributes attributes) {
        super(toForgeProperties(attributes));
        this.attributes = attributes;
        this.forgeType = Suppliers.memoize(() -> cachedFluidType.get(attributes));
    }

    private static Properties toForgeProperties(ArchitecturyFluidAttributes attributes) {
        Properties forge = cachedProperties.computeIfAbsent(attributes, (_a) -> {
            FluidType ft = cachedFluidType.computeIfAbsent(attributes,
                (_aa) -> new ArchitecturyFluidAttributesForge(FluidType.Properties.create(), attributes.getSourceFluid(),
                        attributes));
            return new Properties(Suppliers.memoize(() -> ft), attributes::getSourceFluid, attributes::getFlowingFluid);
        });
        forge.slopeFindDistance(attributes.getSlopeFindDistance());
        forge.levelDecreasePerBlock(attributes.getDropOff());
        forge.bucket(() -> MoreObjects.firstNonNull(attributes.getBucketItem(), Items.AIR));
        forge.tickRate(attributes.getTickDelay());
        forge.explosionResistance(attributes.getExplosionResistance());
        forge.block(() -> MoreObjects.firstNonNull(attributes.getBlock(), (LiquidBlock) Blocks.WATER));
        return forge;
    }

```

Thank you!

digitalseraphim:
…ame ArchitecturyFluidAttributes object is passed in.

This fixes https://github.com/architectury/architectury-api/issues/447. This is the smallest change I could come up with. The only other way I could see to fix it would be to create a Forge  specific version of the SimpleArchitecturyFluidAttributes class that contained a `toForgeType` or `getForgeType`, that cached the value itself, but you'd have to cast it, because the interface type is used for the AFF constructor, and we don't want Forge specific types in the interface, unless we also make a forge specific version of the interface, and that just seems to be getting out of control.. I think this is a nice clean solution to the issue.

### DIFF
--- a/forge/src/main/java/dev/architectury/core/fluid/forge/imitator/ArchitecturyFlowingFluid.java
+++ b/forge/src/main/java/dev/architectury/core/fluid/forge/imitator/ArchitecturyFlowingFluid.java
@@ -42,25 +42,28 @@ import net.minecraftforge.fluids.FluidType;
 import net.minecraftforge.fluids.ForgeFlowingFluid;
 import org.jetbrains.annotations.NotNull;
 
+import java.util.HashMap;
+import java.util.Map;
 import java.util.Optional;
-import java.util.function.Supplier;
 
 public abstract class ArchitecturyFlowingFluid extends ForgeFlowingFluid {
     private final ArchitecturyFluidAttributes attributes;
-    private final Supplier<FluidType> forgeType;
+    private static final Map<ArchitecturyFluidAttributes, Properties> cachedProperties;
+    static {
+        cachedProperties = new HashMap<>();
+    }
     
     ArchitecturyFlowingFluid(ArchitecturyFluidAttributes attributes) {
         super(toForgeProperties(attributes));
         this.attributes = attributes;
-        this.forgeType = Suppliers.memoize(() -> {
-            return new ArchitecturyFluidAttributesForge(FluidType.Properties.create(), this, attributes);
-        });
     }
     
     private static Properties toForgeProperties(ArchitecturyFluidAttributes attributes) {
-        Properties forge = new Properties(Suppliers.memoize(() -> {
-            return new ArchitecturyFluidAttributesForge(FluidType.Properties.create(), attributes.getSourceFluid(), attributes);
-        }), attributes::getSourceFluid, attributes::getFlowingFluid);
+        Properties forge = cachedProperties.computeIfAbsent(attributes, (_a) -> {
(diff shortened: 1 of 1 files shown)


shedaniel:
Does `toForgeProperties` also need to return the same value per the same attribute? If not, this PR can probably be simplified as follows:

image

digitalseraphim:
Dealing with a power outage, doing this on my phone, so please pardon any
shortcuts/typos.

There's a little cleanup here as well, as AFF doesn't need to override
getFluidType, as the Forge implementation returns what it gets from the
properties object returned by toForgeProperties. So there's also no need
for the forgeType member. Instead of creating multiple FluidType instances
that are just going to be thrown away, (as well as Properties objects),
doesn't this make more sense?

shedaniel:
That does make sense, thank you. I will clean this a little bit up (just small nitpicks) and this should be good to go!

shedaniel (approved):

PR #472, opened 2024-01-09
1.20.4 PR merges

MaxNeedsSnacks:


shedaniel (changes requested):
I cannot review each PR if PRs are merged to an intermediary branch like this.

PRs should also indivually be first merged to the lowest common version, 1.20.4 has very low adoption right now and lower versions benefit more people, merging up versions instead of cherry-picking makes things much easier to handle.

PRs will have to be made again (I am sorry max)

PR #493, opened 2024-04-22
Port to 1.20.5

Jab125:


Jab125:
More things, I've removed the `modmenu:api` field and replaced it with `"modmenu": { "badges": ["library"]}`, and I've set the Forge version to `50.0.0` for now, as well as adding Mod Menu to the test mod.

testmod-fabric/build.gradle
```diff
@@ -21,10 +21,16 @@ configurations {
...
 
+repositories {
+    maven { url "https://maven.terraformersmc.com/releases/" }
+}
+
 dependencies {
     modImplementation "net.fabricmc:fabric-loader:${rootProject.fabric_loader_version}"
     modImplementation "net.fabricmc.fabric-api:fabric-api:${rootProject.fabric_api_version}"
 
+    modLocalRuntime("com.terraformersmc:modmenu:${rootProject.mod_menu_version}") { transitive false }
```
shedaniel: I would not add modmenu as a runtime dep because they usually port later than us and their api is pretty much never broken anyways, so I just expect it to work

shedaniel (changes requested):

testmod-fabric/build.gradle
Jab125: I've removed it from the testmod now

shedaniel (approved):

testmod-fabric/build.gradle
```diff
@@ -21,6 +21,10 @@ configurations {
     developmentFabric.extendsFrom common
 }
 
+repositories {
```
shedaniel: Actually remove this too

Jab125: Done

shedaniel:
Thanks for your work! Really saved a lot of time with this version!

PR #494, opened 2024-04-27
Port to 1.20.6

Jab125:
- [x] Fabric
- [x] NeoForge

Should be a very quick port.


shedaniel:
Do we need a port? Doesn't it just works?

Jab125:
> Do we need a port? Doesn't it just works?

We'll have to port in the future anyways, since I doubt NeoForge would keep maintaining 20.5 after 1.20.6 is released.

shedaniel:
Yeah don't worry about that for now, we/I can do that when that's needed, we do that for older similar versions too

Jab125:
I've also fixed a bug with `void acceptAfter(ItemStack after, ItemStack stack, CreativeModeTab.TabVisibility visibility);` and `void acceptBefore(ItemStack before, ItemStack stack, CreativeModeTab.TabVisibility visibility);` where on NeoForge, it was unable to actually place items in the correct locations.

Jab125:
Also requesting the creation of a `1.20.6` branch to merge this into.

shedaniel (approved):

PR #503, opened 2024-05-04
Fix FluidStack.EMPTY always being null

### LINKED ISSUE #500: 1.20.6 FluidStack.empty() returns null (and FluidStack.EMPTY is null)
The static init for `FluidStack.EMPTY` appears to be incorrect:
```java
private static final FluidStack EMPTY = create(Fluids.EMPTY, 0);
```

Following the `create()` method call chain leads to this method:
```java
public static FluidStack create(Fluid fluid, long amount, DataComponentPatch patch) {
    if (fluid == Fluids.EMPTY || amount <= 0) return empty();
    return create(() -> fluid, amount, patch);
}
```

Since the amount is 0, `empty()` is called, which returns `EMPTY`, which at this point is still null.

Jab125:
Closes #500 
My fix adds a null check to `FluidStack.EMPTY` to see if it is null as shown below:
```diff
--- a/common/src/main/java/dev/architectury/fluid/FluidStack.java
+++ b/common/src/main/java/dev/architectury/fluid/FluidStack.java
@@ -43,7 +43,7 @@
 
 public final class FluidStack implements DataComponentHolder {
     private static final FluidStackAdapter<Object> ADAPTER = adapt(FluidStack::getValue, FluidStack::new);
-    private static final FluidStack EMPTY = create(Fluids.EMPTY, 0);
+    private static final FluidStack EMPTY = new FluidStack(() -> Fluids.EMPTY, 0, DataComponentPatch.EMPTY);
     public static final Codec<FluidStack> CODEC = ADAPTER.codec();
     public static final StreamCodec<RegistryFriendlyByteBuf, FluidStack> STREAM_CODEC = ADAPTER.streamCodec();
 ```

### DIFF
--- a/common/src/main/java/dev/architectury/fluid/FluidStack.java
+++ b/common/src/main/java/dev/architectury/fluid/FluidStack.java
@@ -43,7 +43,7 @@ import java.util.function.UnaryOperator;
 
 public final class FluidStack implements DataComponentHolder {
     private static final FluidStackAdapter<Object> ADAPTER = adapt(FluidStack::getValue, FluidStack::new);
-    private static final FluidStack EMPTY = create(Fluids.EMPTY, 0);
+    private static final FluidStack EMPTY = new FluidStack(() -> Fluids.EMPTY, 0, DataComponentPatch.EMPTY);
     public static final Codec<FluidStack> CODEC = ADAPTER.codec();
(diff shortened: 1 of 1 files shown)


shedaniel:
No but, the proper fix would be to make EMPTY not null

shedaniel (approved):

PR #506, opened 2024-05-22
Add Codecs to Registrar

Waterpicker:
I noticed that registries provide a Codec<T> and Codec<Holder<T>. Since a Registry is wrapped by a Registrar, why not have a Registrar provide the same codecs?


shedaniel (approved):

Waterpicker:
So any reason why this hasn't been merged? Just came back to look and its been ready for a long time now.

PR #534, opened 2024-07-07
[1.21] Use RegistryFriendlyByteBuf for extended menus

Urkaz:
Extended menus use `FriendlyByteBuf`, so they can't be used with StreamCodecs to serialize things as they require a `RegistryFriendlyByteBuf`.

Forge and NeoForge already use `RegistryFriendlyByteBuf` natively when calling `player.openMenu`, but Fabric's implementation is more custom on the Architectury side and it creates a `FriendlyByteBuf`.

This PR changes all `FriendlyByteBuf` into `RegistryFriendlyByteBuf` in all places I've seen it related to extended menus.

### DIFF
--- a/common/src/main/java/dev/architectury/registry/menu/ExtendedMenuProvider.java
+++ b/common/src/main/java/dev/architectury/registry/menu/ExtendedMenuProvider.java
@@ -19,9 +19,9 @@
 
 package dev.architectury.registry.menu;
 
-import net.minecraft.network.FriendlyByteBuf;
+import net.minecraft.network.RegistryFriendlyByteBuf;
 import net.minecraft.world.MenuProvider;
 
 public interface ExtendedMenuProvider extends MenuProvider {
-    void saveExtraData(FriendlyByteBuf buf);
+    void saveExtraData(RegistryFriendlyByteBuf buf);
 }
--- a/common/src/main/java/dev/architectury/registry/menu/MenuRegistry.java
+++ b/common/src/main/java/dev/architectury/registry/menu/MenuRegistry.java
@@ -25,6 +25,7 @@ import net.fabricmc.api.Environment;
 import net.minecraft.client.gui.screens.Screen;
 import net.minecraft.client.gui.screens.inventory.MenuAccess;
 import net.minecraft.network.FriendlyByteBuf;
+import net.minecraft.network.RegistryFriendlyByteBuf;
 import net.minecraft.network.chat.Component;
 import net.minecraft.server.level.ServerPlayer;
 import net.minecraft.world.MenuProvider;
@@ -50,10 +51,10 @@ public final class MenuRegistry {
      * @param provider  The {@link MenuProvider} that provides the menu
      * @param bufWriter That writer that sends extra data for {@link MenuType} created with {@link MenuRegistry#ofExtended(ExtendedMenuTypeFactory)}
      */
-    public static void openExtendedMenu(ServerPlayer player, MenuProvider provider, Consumer<FriendlyByteBuf> bufWriter) {
+    public static void openExtendedMenu(ServerPlayer player, MenuProvider provider, Consumer<RegistryFriendlyByteBuf> bufWriter) {
         openExtendedMenu(player, new ExtendedMenuProvider() {
             @Override
-            public void saveExtraData(FriendlyByteBuf buf) {
+            public void saveExtraData(RegistryFriendlyByteBuf buf) {
                 bufWriter.accept(buf);
             }
             
@@ -183,6 +184,6 @@ public final class MenuRegistry {
          * @param buf       The {@link FriendlyByteBuf} for the menu to provide extra data
          * @return A new {@link T} that extends {@link AbstractContainerMenu}
          */
-        T create(int id, Inventory inventory, FriendlyByteBuf buf);
+        T create(int id, Inventory inventory, RegistryFriendlyByteBuf buf);
     }
 }
--- a/fabric/src/main/java/dev/architectury/registry/menu/fabric/MenuRegistryImpl.java
+++ b/fabric/src/main/java/dev/architectury/registry/menu/fabric/MenuRegistryImpl.java
@@ -27,13 +27,13 @@ import io.netty.buffer.ByteBufUtil;
 import io.netty.buffer.Unpooled;
 import net.fabricmc.api.EnvType;
 import net.fabricmc.api.Environment;
-import net.fabricmc.fabric.api.networking.v1.PacketByteBufs;
 import net.fabricmc.fabric.api.screenhandler.v1.ExtendedScreenHandlerFactory;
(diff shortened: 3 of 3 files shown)


MaxNeedsSnacks (approved):

Jab125 (approved):
Would this cause any breaking changes? Since if other mods are already using this, then they may have to recompile their mod.

Urkaz:
> Would this cause any breaking changes? Since if other mods are already using this, then they may have to recompile their mod.

I'm not sure if this will be a breaking change, but I don't know exactly how to check it to confirm it or not 😅 

What I can say for sure is that `RegistryFriendlyByteBuf` extends from `FriendlyByteBuf` so all functions (writeInt, writeBoolean, writeBlockPos, etc) should still be there to use as if it were a `FriendlyByteBuf`.

If the type is left as a `FriendlyByteBuf` in [ExtendedMenuProvider.java](https://github.com/architectury/architectury-api/pull/534/files#diff-13850ce2aef713b248a06ad80356752c428ce88b69088718e4613afcc9da0f6d) and [MenuRegistry.java](https://github.com/architectury/architectury-api/pull/534/files#diff-00af5bad17bdd04dc4e4cac5df7076a31652c4869f5690de69c2d596d0e84c08), then we can cast the `FriendlyByteBuf` to `RegistryFriendlyByteBuf` and use it in StreamCodecs like this:

```java
// Overriding the ExtendedMenuProvider
@Override
public void saveExtraData(FriendlyByteBuf buffer) {
    RegistryFriendlyByteBuf b2 = (RegistryFriendlyByteBuf) buffer;
    ItemStack itemInHand = ItemHelper.getItemInHand(ownerPlayer);
    ItemStack.STREAM_CODEC.encode(b2, itemInHand);
}

// During registration
public static final RegistrySupplier<MenuType<MyMenu>> MY_MENU = MENU_TYPES.register("my_menu", () -> MenuRegistry.ofExtended((id, inventory, buffer) -> {
    RegistryFriendlyByteBuf b2 = (RegistryFriendlyByteBuf) buffer;
    ItemStack itemInHand = ItemStack.STREAM_CODEC.decode(b2);
    return new MyMenu(id, inventory, itemInHand);
}));
```
This example works, but that will still require the proposed changes in [fabric/MenuRegistryImpl.java](https://github.com/architectury/architectury-api/pull/534/files#diff-d2b135397bb9578ef9d6ae403849a9cb1f5200cc27e7ce85b133e70390be6a26) so the same type of buffer is returned by all mod loaders (Forge and NeoForge already use `RegistryFriendlyByteBuf` in their implementation), or fabric will throw a ClassCastException because of the invalid cast.

Edit: I forgot to mention that this change may be required in 1.20.6 too, as I think the codec and buffer changes were made in that version, but I may be wrong.

Jab125:
> > Would this cause any breaking changes? Since if other mods are already using this, then they may have to recompile their mod.
> 
> I'm not sure if this will be a breaking change, but I don't know exactly how to check it to confirm it or not 😅
> 
> What I can say for sure is that `RegistryFriendlyByteBuf` extends from `FriendlyByteBuf` so all functions (writeInt, writeBoolean, writeBlockPos, etc) should still be there to use as if it were a `FriendlyByteBuf`.
> 
> If the type is left as a `FriendlyByteBuf` in [ExtendedMenuProvider.java](https://github.com/architectury/architectury-api/pull/534/files#diff-13850ce2aef713b248a06ad80356752c428ce88b69088718e4613afcc9da0f6d) and [MenuRegistry.java](https://github.com/architectury/architectury-api/pull/534/files#diff-00af5bad17bdd04dc4e4cac5df7076a31652c4869f5690de69c2d596d0e84c08), then we can cast the `FriendlyByteBuf` to `RegistryFriendlyByteBuf` and use it in StreamCodecs like this:
> 
> ```java
> // Overriding the ExtendedMenuProvider
> @Override
> public void saveExtraData(FriendlyByteBuf buffer) {
>     RegistryFriendlyByteBuf b2 = (RegistryFriendlyByteBuf) buffer;
>     ItemStack itemInHand = ItemHelper.getItemInHand(ownerPlayer);
>     ItemStack.STREAM_CODEC.encode(b2, itemInHand);
> }
> 
> // During registration
> public static final RegistrySupplier<MenuType<MyMenu>> MY_MENU = MENU_TYPES.register("my_menu", () -> MenuRegistry.ofExtended((id, inventory, buffer) -> {
>     RegistryFriendlyByteBuf b2 = (RegistryFriendlyByteBuf) buffer;
>     ItemStack itemInHand = ItemStack.STREAM_CODEC.decode(b2);
>     return new MyMenu(id, inventory, itemInHand);
> }));
> ```
> 
> This example works, but that will still require the proposed changes in [fabric/MenuRegistryImpl.java](https://github.com/architectury/architectury-api/pull/534/files#diff-d2b135397bb9578ef9d6ae403849a9cb1f5200cc27e7ce85b133e70390be6a26) so the same type of buffer is returned by all mod loaders (Forge and NeoForge already use `RegistryFriendlyByteBuf` in their implementation), or fabric will throw a ClassCastException because of the invalid cast.
> 
> Edit: I forgot to mention that this change may be required in 1.20.6 too, as I think the codec and buffer changes were made in that version, but I may be wrong.

all `RegistryFriendlyByteBuf`s are `FriendlyByteBuf`s, but not all `FriendlyByteBuf`s are `RegistryFriendlyByteBuf`s, so there may be a problem on fabric.

Also, in Java bytecode, both a method's name _and_ signature are used to figure out what method to call:
```
saveExtraData(Lnet/minecraft/network/FriendlyByteBuf;)V
```
so this would probably break all mods that currently use this.

Urkaz:
I did a few tests on my side and I can confirm that only merging the changes from [fabric/MenuRegistryImpl.java](https://github.com/architectury/architectury-api/pull/534/files#diff-d2b135397bb9578ef9d6ae403849a9cb1f5200cc27e7ce85b133e70390be6a26) doesn't break mods and will continue to work without recompiling. Merging the changes in the other two files ([MenuRegistry.java](https://github.com/architectury/architectury-api/pull/534/files#diff-00af5bad17bdd04dc4e4cac5df7076a31652c4869f5690de69c2d596d0e84c08) and [ExtendedMenuProvider.java](https://github.com/architectury/architectury-api/pull/534/files#diff-13850ce2aef713b248a06ad80356752c428ce88b69088718e4613afcc9da0f6d)) breaks them.

But as I mentioned, those changes are still necessary to fix the custom Architectury's implementation of extended menus for Fabric and make it behave like Forge and NeoForge.

Should I edit this PR with something or is it OK the way it is right now? I'm just asking in case something is blocking it from being merged :)
Thanks!

shedaniel:
I don't think we can afford breaking changes right now, for now you can get the RegistryAccess from the ServerPlayer and create the RegistryFriendlyByteBuf in the consumer yourself, we can merge this change for the next major version so keep this PR open for now.

Urkaz:
Closing as it is no longer needed in recent versions.

PR #596, opened 2025-02-01
Create zh_cn.json and add Chinese MOD instructions

Q2297045667:


Q2297045667:
@shedaniel Can you merge it?

shedaniel:
I don't think there is any use for this translation, it is a test mod for development and we don't need translations for it...

Q2297045667:
Alright

PR #603, opened 2025-02-16
Use new NeoForge reload listeners

rosellle:
This change breaks mods that were using a null listenerId when calling `ReloadListenerRegistry.register(...)`, as NeoForge now requires that all listeners are named.

Summary of changes:
1. The ReloadListenerRegistry.register(PackType type, PreparableReloadListener listener) method has been removed.
2. For the remaining register(...) methods, the `ResourceLocation listenerId` is no longer `@Nullable`.
3. The `ReloadListenerRegistryImpl` for NeoForge now uses the appropriate events.
4. Architectury API now depends on a minimum version of `21.4.84-beta`.

See https://github.com/neoforged/NeoForge/pull/1915 for more information.


shedaniel (approved):

shedaniel:
Thanks!

PR #616, opened 2025-03-23
Port to 1.21.5 rc1

JPeisach:
1.21.5-pre3 had a change to the target for the mixin in MixinResultSlot, which would cause the game to immediately crash.

Then bump to 1.21.5-rc1, no changes needed. Leaving it at Fabric API 0.119.3+1.21.5, as 0.119.4 includes breaking changes.


shedaniel (approved):
Thank you so much!

PR #666, opened 2025-10-01
First efforts towards 1.21.9

ThexXTURBOXx:
Completely untested and does not compile yet...
Maybe this helps anyway!


shedaniel:
Thanks for your efforts! But I have it compiled already but removing Environment and OnlyIn annotation is the part that takes the most time

PR #685, opened 2025-12-10
Port to 1.21.11

pynickle:
Port to 1.21.11.

The primary changes involve adjustments to certain class locations and the renaming of ResourceLocation to Identifier.

Also includes some biome effects values migrated to the datapack, they are no longer needed in the code..

 Also updated some GitHub Action versions.


common/src/main/java/dev/architectury/hooks/level/biome/EffectsProperties.java
```diff
@@ -19,67 +19,31 @@
...
         
-        EffectsProperties.Mutable setWaterFogColor(int color);
-        
-        EffectsProperties.Mutable setSkyColor(int color);
-        
         EffectsProperties.Mutable setFoliageColorOverride(@Nullable Integer colorOverride);
         
+        EffectsProperties.Mutable setDryFoliageColorOverride(@Nullable Integer colorOverride);
+        
         EffectsProperties.Mutable setGrassColorOverride(@Nullable Integer colorOverride);
```
JPeisach, line 45: The implementations for this actually do the same as foliageColorOverride:
```java
@Override
        public Mutable setGrassColorOverride(@Nullable Integer colorOverride) {
            builder.foliageColorOverride = Optional.ofNullable(colorOverride);
            return this;
        }
```

`builder.grassColorOverride` exists. not sure if it should be included in the PR

pynickle: Weird, but l make a fix then.

JPeisach (approved):
looks good to me - ~some things like fog color I'm not sure what the new ways things like fog is supposed to be done but they seem to be moved in main game functions anyways, so it should hopefully be fine~ nevermind as you said its in datapack

PR #691, opened 2026-01-18
Refactor listeners to use List interface

Viola-Siemens:
Changed listeners from ArrayList to List for better abstraction and added type check for trimming listeners.

https://github.com/Viola-Siemens/CME-Suck-My-Duck/issues/2


Viola-Siemens:
By the way, why the workflow is stuck in my repo?
image

Issue #491, opened 2024-04-13
EntityEvent.ENTER_SECTION is not called on Forge


Jab125:
Has been broken since https://github.com/architectury/architectury-api/commit/bf34031711d3861e91d7f498a1d9c3c43c49d9e2

PR #692, opened 2026-01-24
Fix EntityEvent.ENTER_SECTION not being called on NeoForge

st0x0ef:
This event was invoked on Fabric but not on NeoForge. This PR fix that by implementing it correctly for 1.21.11. Tested with `testmod-neoforge` and it seems to work perfectly.


st0x0ef:
fix #491

PR #729, opened 2026-07-16
Add registries as additional parameter to loot table modification event

ThexXTURBOXx:
Alas, this would be a breaking change and additionally it would also affect *every* version of Architectury API since 1.21.
The extra object is important when functions such as `EnchantRandomlyFunction#randomApplicableEnchantment` are to be used.

### DIFF
--- a/common/src/main/java/dev/architectury/event/events/common/LootEvent.java
+++ b/common/src/main/java/dev/architectury/event/events/common/LootEvent.java
@@ -21,6 +21,7 @@ package dev.architectury.event.events.common;
 
 import dev.architectury.event.Event;
 import dev.architectury.event.EventFactory;
+import net.minecraft.core.HolderLookup;
 import net.minecraft.resources.ResourceKey;
 import net.minecraft.world.level.storage.loot.LootPool;
 import net.minecraft.world.level.storage.loot.LootTable;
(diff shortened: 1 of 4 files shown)


common/src/main/java/dev/architectury/event/events/common/LootEvent.java
```diff
@@ -65,12 +66,13 @@ interface ModifyLootTable {
...
-         * @param builtin if {@code true}, the loot table is built-in;
-         *                if {@code false}, it is from a user data pack
+         * @param registries the registries provider
+         * @param key        the loot table key
+         * @param context    the context used to modify the loot table
+         * @param builtin    if {@code true}, the loot table is built-in;
+         *                   if {@code false}, it is from a user data pack
          */
-        void modifyLootTable(ResourceKey<LootTable> key, LootTableModificationContext context, boolean builtin);
+        void modifyLootTable(HolderLookup.Provider registries, ResourceKey<LootTable> key, LootTableModificationContext context, boolean builtin);
```
ThexXTURBOXx: Since this would be a breaking change, please let me know how to properly handle this. Should I make the new function `default` with an empty body and let the old one stay (but deprecate it), for instance?

st0x0ef: We should keep the old one for compatibility, but marked as deprecated (for removal in 26.3).

st0x0ef:
Looks good! Just one small change. Thanks for your help!

common/src/main/java/dev/architectury/event/events/common/LootEvent.java
ThexXTURBOXx: Just like I thought - I have re-added the old entry as deprecated for removal. Thanks!

PR #732, opened 2026-07-20
Probably fix publishing to CurseForge

ThexXTURBOXx:
Don't know 100% whether this fixes the issue with the pipeline, but I am rather certain.
Should be backported to "LTS" releases as well (you mentioned that 1.21[.1] and 26.1 should be getting the fix from #729 as well).


ThexXTURBOXx:
#733 may be better than this PR?

